PR #99, opened 2022-04-08
Implementation of storybook

### LINKED ISSUE #78: Implementation of storybook
- [x] We should able to create independent sections in storybook for documenting general stuff (like how to use grace, some general topics etc)
- [x] Every component should have it's own storybook definitions in a file in its own folder (`component.story.ts`)
- [x] Component storybook should include:
  - [x] Basic documentation about where and how to use this component
  - [x] Reference for the properties and events of this component
  - [x] Demos for the variations of every property and behaviours of the component

muratcorlu:
* Component story format converted to `mdx`. This way component stories are much simpler and clearer.
* Some example documents added for the design system. I used built-in components that Stroybook provides for Color Palette, Typography and Iconography.
* Simplified Storybook configuration and dependencies.
* Some example stories added for `gr-button` component.

I'm working on fixing those:

- [ ] Boolean attributes are having empty string values in code examples. Like `<gr-button block="">Save</gr-button>` It should be same as we write in story file (`<gr-button block>Save</gr-button>`)
- [x] `styles` property of components are visible in auto-generated reference list. It should not mentioned in that list.
- [x] Editable property list should be visible in Canvas tab to play with properties.
- [x] Accessibility plugin is not visible.

Closes #78


muratcorlu:
> * Boolean attributes are having empty string values in code examples. Like `<gr-button block="">Save</gr-button>` It should be same as we write in story file (`<gr-button block>Save</gr-button>`)

I couldn't find a solution for this item. I mentioned this issue in multiple slack workspaces that I'm involved but couldn't reveice any suggestion. Also [asked on StackOverflow](https://stackoverflow.com/questions/71999979/storybook-web-component-code-examples-extra-comment-tags). No answer yet.

muratcorlu:
I'm merging this without a solution for code example issues. We'll check that topic again soon.

PR #100, opened 2022-04-17
Linting rules and making it part of workflow

leventozen:
- Added `tsc`, `eslint`, `prettier` and `stylelint` as linters.
- Set rules to run lint `fix` at `pre-commit` and lint `check` at `pull request` by using `husky` and `Github Actions.`
- Removed unnecessary `postcss` dependencies.
- Simplified dependencies.

Right now, I am working on:
- [ ] _removing `postcss-lit` from `stylelintrc` to have better development experience._ - **will be evaluated after bundling task**
- [x] creating documentation for linter configs.

Closes https://github.com/Trendyol/grace/issues/77


leventozen:
I found a way to be able to solve this `removing postcss-lit from stylelintrc to have better development experience.` task by creating a script that generate lit css format from plain css. However, I think it is not a good idea before evaluating our way of bundling. Therefore, this will be postponed after we start to work on bundling mechanism task.

muratcorlu (approved):

package.json
```diff
@@ -107,6 +107,7 @@
     "eslint-config-prettier": "^8.5.0",
     "eslint-plugin-import": "^2.26.0",
     "gaze": "^1.1.3",
+    "html-webpack-plugin": "^5.5.0",
```
muratcorlu, line 110: Do we need this? I couldn't see any reference to this plugin in this PR.

package.json
```diff
@@ -120,14 +125,22 @@
     "rollup-plugin-terser": "^7.0.2",
     "rollup-plugin-typescript2": "^0.31.2",
     "rsync": "^0.6.1",
+    "storybook": "^6.4.19",
+    "stylelint": "^14.6.1",
+    "stylelint-config-standard": "^25.0.0",
     "svgo": "^2.7",
     "ts-lit-plugin": "^1.2.1",
-    "typescript": "^4.6.3",
+    "typescript": "^4.0.0",
```
muratcorlu, line 133: Do we need to reduce TS version?

package.json
leventozen: Possible merging problem, I will fix in the next commit.

PR #118, opened 2022-05-29
feat: button component

leventozen:
Button component implemented. It's documentation with more detailed information will be provided in different PR.

Closes https://github.com/Trendyol/grace/issues/118


src/components/gr-button/gr-button.css
```diff
@@ -1,78 +1,246 @@
...
+:host([link]) .button {
+  border: none;
+  background: none;
+  text-decoration: underline;
+
+  --gr-content-color: var(--gr-color-content-outline-primary);
+  --gr-button-font: var(--gr-font-title-3-medium);
+}
+
+:host([link]:hover:not([disabled])) .button {
```
muratcorlu, line 64: I think we can simplify this `:hover` styles by defining new variables for hover styles and setting them with variants. Like:

```css
.button {
    …
    —bl-button-hover-color: {primaryHoverColor};
}

.button:hover {
     background-color: var(—bl-button-hover-color);
}

:host([disabled]) .button {
     background-color: {disabledButtonBgColor};
}

:host([secondary]) {
   …
   —bl-button-hover-color: {secondaryHoverColor};
}
```

I expect this can reduce clutter a lot in CSS file.

src/components/gr-button/gr-button.ts
```diff
@@ -1,55 +1,129 @@
...
+export type TargetType = '_blank' | '_parent' | '_self' | '_top';
+
 @customElement('gr-button')
 export default class GrButton extends LitElement {
   static get styles(): CSSResultGroup {
     return [style];
   }
 
-  @state() private hasFocus = false;
+  @property({ type: String, attribute: 'size' })
```
muratcorlu, line 39: Do we need `attribute` option here?

Also since we use size attribute selector in our styles, we need to use `reflect: true` to be sure that we’ll always have `size` attribute even if property set via JS.

Same applies for other attributes that we use in CSS (like `secondary`, `success`, `link`, `outline` etc.) We need `@property` definitions for them to make it possible to set them as component DOM property, not just as an HTML attribute.

src/themes/default.css
```diff
@@ -4,14 +4,30 @@
...
+  --gr-color-primary-outline: #FFF;
   --gr-color-secondary: #273142;
   --gr-color-secondary-hover: #0f131a;
-  --gr-color-success: #0bc15c;
-  --gr-color-success-hover: #09a44e;
-  --gr-color-content-primary: #273142;
-  --gr-color-content-secondary: #6e7787;
-  --gr-color-content-tertiary: #95a1b5;
-  --gr-color-content-passive: #afbbca;
+  --gr-color-secondary-outline: #FFF;
```
muratcorlu, line 10: Outline buttons has always `transparent` backgrounds. So I think we should not need those variables.

src/themes/default.css
```diff
@@ -4,14 +4,30 @@
...
+  --gr-color-tertiary-hover: #D5D9E1;
+  --gr-color-tertiary-outline: #FFF;
+  --gr-color-success: #0BC15C;
+  --gr-color-success-hover: #09A44E;
+  --gr-color-success-outline: #FFF;
+  --gr-color-error: #FF5043;
+  --gr-color-error-hover: #FF3028;
+  --gr-color-error-outline: #FFF;
+  --gr-color-content-primary: #FFF;
+  --gr-color-content-outline-primary: #f27a1a;
```
muratcorlu, line 21: Outline font colors should be always same with the main color of the variant. So I think we should not need those outline color variables.

src/themes/default.css
```diff
@@ -4,14 +4,30 @@
...
+  --gr-color-content-secondary: #FFF;
+  --gr-color-content-outline-secondary: #273142;
+  --gr-color-content-tertiary: #273142;
```
muratcorlu, line 24: Font colors are always either white or same as the variant color, except tertiary. Let’s discuss this inconsistency in tertiary button with our design team. If we can align tertiary with others, we should not need content color definitions for buttons.

leventozen: Yeah, it is known issue by me. I already contacted to design team. For now, I will extract tertiary button for the design system's sake. You can check it in the next commit.

src/themes/default.css
leventozen: Known issue. It will be solved in the next commit 👍

src/components/gr-button/gr-button.ts
leventozen: That's an interesting point. I will look at it.

src/themes/default.css
leventozen: Known issue. It will be solved in the next commit 👍. PR will be ready for review on Monday.

leventozen:
@buseselvi, we should help user about answering these questions: what for? why? how? when to use?, etc.
Here is the link of our button component that you can take a look. We should provide more details about our variants, etc.
[button-component](https://6267ebb4dd54f5004a65f14a-ywchmugrif.chromatic.com/?path=/docs/components-button--primary-button)
Good luck! 👋 
PS: I extracted the tertiary button for now. After we make progress about it like having answers for the questions above, I will add.

src/themes/default.css
leventozen: You can follow this [comment](https://github.com/Trendyol/grace/pull/118#issuecomment-1146889597) for latest news about **tertiary** button.

muratcorlu (approved):

buseselvi (approved):

PR #121, opened 2022-06-01
feat(icon): arrangement icon is added

buseselvi:


muratcorlu:
Are we sure with the name as `arrangement`?

PR #129, opened 2022-06-21
feat: input component

muratcorlu:
Known issues:

- [ ] Storybook demos for validation and focus states don't have a proper visual regression tests
- [ ] Component argument reference has duplicate values for the properties like `help-text` since its property is `helpText`. We need to analyze how to properly document properties and attributes.

Also there are many missing features that we'll need near future for sure. But we wanted to iterate on this because input component is a quite complex component and we thought it makes sense to improve it step by step.


muratcorlu:
@buseselvi can you please make a UI review on Chromatic?

buseselvi:
The font and font size of the input labels look different from the design, can we fix it?

src/components/input/bl-input.css
```diff
@@ -0,0 +1,159 @@
...
+
+input {
+  outline: none;
+  box-sizing: border-box;
+  border: none;
+  width: 100%;
+  font: var(--bl-font-title-3-regular);
+  padding: var(--bl-input-padding-vertical) var(--bl-input-padding-horizontal);
+  margin: 0;
+  box-shadow: inset 0 0 0 1px var(--bl-input-border-color);
```
leventozen, line 21: is box-shadow necessary?

leventozen, line 21: is box-shadow necessary?

package.json
```diff
@@ -78,15 +78,15 @@
...
+    "@storybook/addon-a11y": "6.5.5",
+    "@storybook/addon-actions": "6.5.5",
+    "@storybook/addon-essentials": "6.5.5",
+    "@storybook/addon-links": "6.5.5",
+    "@storybook/addon-storysource": "6.5.5",
+    "@storybook/web-components": "6.5.5",
     "@typescript-eslint/eslint-plugin": "^5.18.0",
     "@typescript-eslint/parser": "^5.18.0",
-    "@web/dev-server-esbuild": "^0.3.0",
+    "@web/dev-server-esbuild": "0.2.16",
```
leventozen, line 89: why did we need to downgrade the es dev server version?

src/components/input/bl-input.css
muratcorlu: Not anymore, I just changed them to borders. But as I mentioned in [my previous comment](https://github.com/Trendyol/baklava/pull/129#issuecomment-1171220747) we just changed the approach about setting the height of the components.

leventozen:
> > Medium input height looks 32.5px, should be 32px. Also, Large input height looks 40.5px, should be 40px.
> 
> We were using line-height for calculating input height but while we are implementing `bl-badge` (#131) we converted values to `rem` and decided to starting setting `height`. Once that PR merged, we needed to update current components (including `button` @leventozen ) to use this new approach.
> 
> I fixed it and should be ok now.



> > Medium input height looks 32.5px, should be 32px. Also, Large input height looks 40.5px, should be 40px.
> 
> We were using line-height for calculating input height but while we are implementing `bl-badge` (#131) we converted values to `rem` and decided to starting setting `height`. Once that PR merged, we needed to update current components (including `button` @leventozen ) to use this new approach.
> 
> I fixed it and should be ok now.

Yes, I am aware of it. I commented for this reason actually. I checked, looks okay for me too.👍

buseselvi:
When an error message appears in the input validation section in storybook, the other input slides down, can we fix it so that it stays above? According to the descriptions, the alignment of the adjacent inputs should not shift.

Ekran Resmi 2022-06-30 17 52 26

muratcorlu:
> When error message appears in the input validation section in storybook, the other input slides down, can we fix it so that it stays above? When we show descriptions, the alignment of the other inputs should not change.
> 
> Ekran Resmi 2022-06-30 17 52 26

This is fixed.

image

But for doing that, I just styled demo pane to align component demostrations to the top. As a result of this that demo looks not aligned but I think we'll accept that. Right?

image

buseselvi:
If we can't do bottom alignment in this part separately, can we put these 3 input types one under the other? We do not use inputs in different label styles side by side in the design. 

Also, the text color of the input on the right looks different from the others, can we check?

muratcorlu:
@buseselvi Good catch! 😊 Fixed it. Please check this version. Storybook will be re-rendered soon.

image

leventozen (approved):

buseselvi:
Can we make the label background primary background color (white)?

image

buseselvi (approved):

PR #131, opened 2022-06-23
feat: badge component

DamlaDemir:
badge component, new colors implemented and fixed sizes.

Closes https://github.com/Trendyol/grace/issues/130


src/components/badge/bl-badge.css
```diff
@@ -0,0 +1,52 @@
+:host {
+  display: inline-block;
+  max-width: 100%;
+
+  --bl-badge-bg-color: var(--bl-color-accent-primary-background);
+  --bl-badge-color: var(--bl-color-primary);
+  --bl-badge-font: var(--bl-size-2xs);
```
muratcorlu, line 7: I think this value is wrong.

src/components/badge/bl-badge.stories.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+export const SizeTemplate = (args) => html`
+${BadgeTemplate({ size:'small', ...args })}
+${BadgeTemplate({ size:'medium', ...args })}
+${BadgeTemplate({ size:'large', ...args })}
+`;
+
+
+# Badge
+The badge can keep different types of content such as a number, status or icon.
```
muratcorlu, line 53: What does "number" mean here? Is there a use case example for this?

src/components/badge/bl-badge.stories.mdx
```diff
@@ -0,0 +1,106 @@
...
+Default background color is `--bl-color-accent-primary-background` and default color is `--bl-color-primary`. 
+
+<Canvas>
+<Story name="Variants" args={{ content: 'Badge' }}>
+    {VariantTemplate.bind({})}
+  </Story>
+</Canvas>
+
+## Icon Badges
+To boost `UX`, you might want to add icons to your badge.Our icons are defined to be `left` of the default slot.
```
muratcorlu, line 81:
```suggestion
To boost UX, you might want to add icons to your badge. Our icons are defined to be left of the default slot.
```

\`\` styles should be used for code references.

src/components/badge/bl-badge.css
```diff
@@ -0,0 +1,52 @@
...
+  font-kerning: none;
+  height: var(--bl-badge-height);
+}
+
+:host([size='small']) {
+  --bl-badge-font: var(--bl-font-caption-small-text);
+  --bl-badge-height: var(--bl-size-m);
+}
+
+:host([size='medium']) {
```
muratcorlu, line 38: Since `medium` is the default size, I think we should not need to set a style to medium. Instead `--bl-badge-font` should be set in `:host` with default font.

src/components/badge/bl-badge.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      `
+        <span class="badge">
+          <slot name="badge-icon"></slot>
+          <slot></slot>
+        </span>
+      `
+    );
+  });
+
+  it('check default values', async () => {
```
muratcorlu, line 40: I couldn't find a standard document about this but I always tend to write case definitions starts with `should...`. So this could be `it('should have correct default values', ...` Maybe we can agree on a standard and document on our contribution guideline.

src/components/badge/bl-badge.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  render(): TemplateResult {
+    const icon = this.icon && this._hasIconBySize ? html`<bl-icon name=${this.icon} />` : '';
+    const slots = html`<slot name="badge-icon">${icon}</slot> <slot></slot>`;
+    const classes = classMap({
+      badge: true,
+    });
+
+    return html`<span class=${classes}>${slots}</span>`;
```
muratcorlu, line 45: No need to have `classMap` here.

```suggestion
    return html`<span class="badge">${slots}</span>`;
```

src/components/badge/bl-badge.ts
```diff
@@ -0,0 +1,53 @@
...
+  size: BadgeSize = 'medium';
+
+  @property({ type: String })
+  icon?: string;
+
+  get _hasIconBySize() {
+    const hasIconSizes = ['medium', 'large'];
+
+    return hasIconSizes.includes(this.size);
+  }
```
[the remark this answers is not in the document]
muratcorlu: If this is because we don't show icons in `small` size, I would suggest hiding icons for small size in CSS. So we'll not need to check this on every render cycle.

src/components/badge/bl-badge.css
DamlaDemir: you are right this value is wrong i will fix the "--bl--badge-font" value.

src/components/badge/bl-badge.css
DamlaDemir: You are right, this will provide the code more readable.

src/components/badge/bl-badge.test.ts
DamlaDemir: The button tests were written like this so I wrote the same way. I agree with your opinion, we can write like this "should ... when ..".

src/components/badge/bl-badge.stories.mdx
DamlaDemir: No this is not a use case I meant to say we can keep any content number, text, icon etc. If this is confusing, we can delete it.

src/components/badge/bl-badge.test.ts
muratcorlu: I'm not sure if we need given-when-then comments but I liked the description 😊

src/components/badge/bl-badge.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      if (/info\.svg$/.test(url.toString())) {
+        return new Response('<svg></svg>');
+      }
+      return new Response('', { status: 404 });
+    };
+  });
+
+  after(() => {
+    window.fetch = oldFetch;
+  });
```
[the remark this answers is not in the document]
muratcorlu: I created a separate issue for this #147

buseselvi (approved):

Issue #124, opened 2022-06-05
Optimize Github Actions workflows


leventozen:
We are using `npm ci` to install all the dependencies. It uses only package lock file, so npm can skip the first steps (you can find more detailed information [here](https://docs.npmjs.com/cli/v7/commands/npm-ci)). It does clean install, in other words deletes `node_modules`. However, we can cache `~/.npm` to prevent redownloading all this stuff.

muratcorlu:
@leventozen can we consider this done with #152 ?

PR #152, opened 2022-07-04
ci: npm cache and remove playwright action

muratcorlu:
While I was working on implementing sonarqube, I noticed that `setup-node` action has [caching capability](https://github.com/actions/setup-node#caching-global-packages-data). Since we already have #124 I just wanted to try this simple change.

Also I noticed playwright-action [encourages to use playwright-cli](https://github.com/microsoft/playwright-github-action#%EF%B8%8F-you-dont-need-this-github-action-%EF%B8%8F) instead of their action. I followed their suggestion here too.

### DIFF
--- a/.github/workflows/publish-chromatic.yml
+++ b/.github/workflows/publish-chromatic.yml
@@ -12,12 +12,11 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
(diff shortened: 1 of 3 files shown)


muratcorlu:
I can see very little improvement in job times for dependency installation and testing preparation. But in general, doesn't affect all for now.

But, still, I think these are the advised way of using those actions.

PR #153, opened 2022-07-05
chore: prevent using old node versions

muratcorlu:
`package-lock.json` format is changed with Node v16. And it's not compatible with Node v14. We put engine requirement in our package.json file as `>=16` but this only shows a warning and many people don't even notice it. Then we are having contributions from the dev env. both node v14 and v16. That causes a full re-write of `package-lock.json` files in our commits.

This PR prevents this by adding [`engine-strict`](https://docs.npmjs.com/cli/v7/using-npm/config#engine-strict) rule to our `.npmrc` file. After this if you make `npm install` in an environment that older node version than v16, you see an error like below:

```bash
➜  baklava git:(prevent-old-node-versions) ✗ npm i
npm ERR! code ENOTSUP
npm ERR! notsup Unsupported engine for @trendyol/baklava@0.0.0-dev: wanted: {"node":">=16"} (current: {"node":"14.19.3","npm":"6.14.17"})
npm ERR! notsup Not compatible with your version of node/npm: @trendyol/baklava@0.0.0-dev
npm ERR! notsup Not compatible with your version of node/npm: @trendyol/baklava@0.0.0-dev
npm ERR! notsup Required: {"node":">=16"}
npm ERR! notsup Actual:   {"npm":"6.14.17","node":"14.19.3"}

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/muratcorlu/.npm/_logs/2022-07-05T11_51_54_023Z-debug.log
```

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true


leventozen (approved):
At first glance, I didn't see the `engine` definition in the commit, but we already define it in the `package.json`. We may think about adding a rule for npm, too.

PR #155, opened 2022-07-14
test: bl-icon is mocked for tests other than its own test

### LINKED ISSUE #147: Mocking BlIcon component in unit tests
Inside the test of every component that uses `bl-icon`,[ we need to mock fetch method](https://github.com/Trendyol/grace/blob/9b3d3f2179323244d1a8faefb940586c09bc9354/src/components/button/bl-button.test.ts#L13) to bypass `bl-icon`s SVG fetching requests. Since there will be many components that will use `bl-icon`, we need to find a better solution for this.

We can either mock `bl-icon` component completely in all of the tests (preferred solution) or we can write a helper to move this fetch mocking solution to a single place.

muratcorlu:
No more need to require mocking fetch method because of using `bl-icon` component
inside others.

Closes #147

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,6 @@
         "@typescript-eslint/eslint-plugin": "^5.18.0",
         "@typescript-eslint/parser": "^5.18.0",
         "@web/dev-server-esbuild": "0.2.16",
-        "@web/dev-server-import-maps": "^0.0.6",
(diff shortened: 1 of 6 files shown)


leventozen (approved):
Well done!

DamlaDemir (approved):

PR #159, opened 2022-07-19
140 tooltip component

DamlaDemir:
Tooltip component completed.

[issues 140](https://github.com/Trendyol/baklava/issues/140)


muratcorlu:
Something wrong with storybook file. It’s not published with tooltip component and UI tests are not triggered. Can you check it?

src/components/tooltip/bl-tooltip.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+  argTypes={{
+    placement: {
+        control: 'text'
+    }
+  }}
+/>
+
+export const IconTriggerTemplate = (args) => html`
+<bl-tooltip placement="${ifDefined(args.placement)}" style='--bl-tooltip-position:fixed'>
+  <bl-icon slot="tooltip-trigger" name="settings"></bl-icon>
```
muratcorlu, line 19:
```suggestion
  <bl-button icon="settings" text label="Settings"></bl-button>
```
We don't want to promote using `bl-icon` everywhere by itself. Instead -most of the time- people should use icon-only text buttons. So let's give good examples here.

src/components/tooltip/bl-tooltip.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+export const ButtonTriggerTemplate = (args) => html`
+<bl-tooltip placement="${ifDefined(args.placement)}" style='--bl-tooltip-position:fixed'>
+  <bl-button slot="tooltip-trigger" icon="plus_fill"></bl-button>
+  You can add text, photos and graphics.
+</bl-tooltip>`
+
+
+export const TextTriggerTemplate = (args) => html`
+<bl-tooltip placement="${ifDefined(args.placement)}" style='--bl-tooltip-position:fixed'>
+  <span slot="tooltip-trigger">Status</span>
```
muratcorlu, line 32: Text example doesn't look good styling. And @buseselvi confirmed that we can just give examples with only buttons for now.

src/components/tooltip/bl-tooltip.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+  Whether the user account is enabled or disabled.
+</bl-tooltip>`
+
+export const PlacementTemplate = (args) => html`
+<bl-tooltip placement="${ifDefined(args.placement)}" 
+style='--bl-tooltip-position:fixed;margin:5px'>
+  <bl-icon slot="tooltip-trigger" name="help"></bl-icon>
```
muratcorlu, line 39:
```suggestion
  <bl-button icon="help" text label="Help"></bl-button>
```
Same rules apply here.

src/components/tooltip/bl-tooltip.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+Tooltips display informative text when users hover over an element.
+
+## Basic Usage
+
+Tooltip can be used with any trigger such as button, icon, text, etc.
+Only you have to give slot name as `tooltip-trigger`. The remaining content will be shown in the tooltip hovering over the trigger component.
+
+Tooltip position is absolute as default. If your wrapper has `position:relative` and `overflow:hidden` the tooltip is clipped and isn't shown correctly.
+In this case, you should give `position:fixed` as below.
```
muratcorlu, line 77:
```suggestion
Tooltip position is absolute as default. If your wrapper has `position:relative` and `overflow:hidden` the tooltip is clipped and isn't shown correctly.

In this case, you should give `position:fixed` as below.
```

src/components/tooltip/bl-tooltip.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+* `left-start`
+* `left`
+* `left-end`
+* `right-start`
+* `right`
+* `right-end`
+
+If there is not enough room for the tooltip in the given placement, the tooltip is repositioned automatically according to the empty area. 
+For example, if there is not enough room on the top, the tooltip is shown on the bottom.
+
```
muratcorlu, line 112:
```suggestion
If there is not enough room for the tooltip in the given placement, the tooltip is repositioned automatically according to the empty area.

For example, if there is not enough room on the top, the tooltip is shown on the bottom.

```

src/components/tooltip/bl-tooltip.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    //when
+    const el = await fixture<typeOfBlTooltip>(html`<bl-tooltip>Test</bl-tooltip>`);
+
+    //then
+    expect(el.placement).to.equal('top');
+  });
+
+  it('should be rendered with slot', async () => {
+    //when
+    const el = await fixture<typeOfBlTooltip>(html`<bl-tooltip><bl-button slot='tooltip-trigger'>Test</bl-button>
```
muratcorlu, line 46: Button component is not imported in the test. Consider importing it or not using it here.

src/components/tooltip/bl-tooltip.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  private show() {
+    this._visible = true;
+    this.setTooltip();
+  }
+
+  private hide() {
+    this._visible = false;
+  }
```
muratcorlu, line 92: Providing some events like `bl-tooltip-show` and `bl-tooltip-hide` could be useful in some cases.

src/components/tooltip/bl-tooltip.ts
```diff
@@ -0,0 +1,119 @@
...
+        left: 'right',
+      };
+      const tooltipPlacement = placement.split('-')[0] as keyof typeof arrowDirections;
+      const arrowDirection = arrowDirections[tooltipPlacement];
+
+      this.host.style.setProperty('--bl-tooltip-left',`${x}px`);
+      this.host.style.setProperty('--bl-tooltip-top',`${y}px`)
+      this.host.style.setProperty('--bl-tooltip-arrow-left',arrowX);
+      this.host.style.setProperty('--bl-tooltip-arrow-top',arrowY);
+      this.host.style.setProperty(`--bl-tooltip-arrow-${arrowDirection}`,'-4px');
```
muratcorlu, line 81: If placement dynamically changes for same tooltip, this property value can be leftover for next value. For example, when placement is `top`, this will set `--bl-tooltip-arrow-bottom: -4px`. Then if placement becomes `right` this will set `--bl-tooltip-arrow-left: -4px` but ``--bl-tooltip-arrow-bottom` will remain as `-4px`. I don't know if this cause a visual issue but looks like a potential issue.

src/components/tooltip/bl-tooltip.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  private setTooltip() {
+    computePosition(this.trigger, this.tooltip, {
+      placement: this.placement,
+      strategy: this._position,
+      middleware: [offset(8), shift({ padding: 5 }), flip(),arrow({ element: this.arrow,padding: 5})],
+    }).then(({ x, y, placement, middlewareData }) => {
+      const arrowX = middlewareData.arrow?.x != null ? `${middlewareData.arrow?.x}px` : ' ';
+      const arrowY = middlewareData.arrow?.y != null ? `${middlewareData.arrow?.y}px`  : ' ';
```
muratcorlu, line 67: Is there a chance to have a `null` value for those values? If so what happens when we set empty string for `--bl-tooltip-arrow-left` ?

muratcorlu (commented):
Good work! I have some remarks.

src/components/tooltip/bl-tooltip.ts
DamlaDemir: This scenario may happen in automatic placement when scrolling the page. When I check I couldn't see any visual problem but we can set `--bl-tooltip-arrow-bottom: 0 `and `--bl-tooltip-arrow-right: 0` to ensure.

src/components/tooltip/bl-tooltip.ts
DamlaDemir: The arrow middleware is designed only to position the arrow on one axis. If the tooltip placement is `top` or `bottom` it gives only `x` position and if the tooltip placement is `right` or `left` it gives only `y` position. 

We set an empty string because we don't want to define `tooltip-arrow-left` when the placement is `right` or `left`. 
And we don't want to define `tooltip-arrow-top` when the placement is `top` or `bottom`.

muratcorlu (approved):
Good job!

buseselvi (approved):

PR #181, opened 2022-08-01
feat: add react support

baspinarenes:


docs/using-baklava-in-react.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+## Usage
+
+Simple usage is as follows:
+
+```javascript
+import { BlButton } from "@trendyol/baklava/dist/react";
+import { setIconPath } from "@trendyol/baklava";
+import "@trendyol/baklava/dist/themes/default.css";
+
+setIconPath("https://cdn.jsdelivr.net/npm/@trendyol/baklava@beta/dist/assets");
```
muratcorlu, line 55: These are not needed when we use via CDN, right? People can use React from NPM and other parts from CDN, I assume.

scripts/generate-react-exports.js
```diff
@@ -0,0 +1,109 @@
...
+      const {
+        groups: { componentName, fileName },
+      } = line.match(/as (?<componentName>.*) } from ['`"](?<path>.*\/(?<fileName>[\w-]*))['`"];$/);
+
+      componentNameDict[componentName] = fileName;
+    } catch (error) {}
+  }
+
+  return componentNameDict;
+}
```
muratcorlu, line 19: We generate `custom-elements.json` file and put to dist folder with `npm run analyze` command. We can use that json file to read component list and their properties etc. I think this can simplify process here a lot.

scripts/generate-react-exports.js
```diff
@@ -0,0 +1,109 @@
...
+    if (litEventName) {
+      const reactEventName = `on${slicedLitEvent[0].toUpperCase()}${slicedLitEvent.slice(1)}`;
+      eventDict[reactEventName] = litEventName;
+    } else {
+      eventDict[eventMatch.groups.reactEventName] = eventMatch.groups.eventName;
+    }
+  }
+
+  return eventDict;
+}
```
muratcorlu, line 70: `custom-elements.json` file also includes events of the components.

scripts/build.js
```diff
@@ -14,7 +14,8 @@ const args = parseArgs(process.argv.slice(2), {
   try {
     const buildOptions = {
       entryPoints: [
-        'src/baklava.ts',
+        'src/index.ts',
+        'src/react/index.ts',
```
muratcorlu, line 18: Do we need to rename main file? Since our main usage approach is CDN, using library/component name as file name has some practical advantages. In the developer tools network tab, you can see more informational `baklava.js` instead of `index.js`. So I would prefer to keep `baklava.ts` and add `baklava-react.ts` instead of `react/index.ts`.

src/react/index.ts
```diff
@@ -0,0 +1,30 @@
...
+const _BlTooltip = createComponent(React, "bl-tooltip", BlTooltip, {
+  onShow: "bl-tooltip-show",
+  onHide: "bl-tooltip-hide",
+});
+
+export { _BlIcon as BlIcon };
+export { _BlButton as BlButton };
+export { _BlInput as BlInput };
+export { _BlBadge as BlBadge };
+export { _BlTooltip as BlTooltip };
```
muratcorlu, line 30: Since this is an autogenerated file, it should be ignored from git.

muratcorlu (changes requested):
Good job folks! I have some remarks.

muratcorlu:
Pipeline fails because of linter. Please run `npm run lint` before committing your work.

We need to run linter in our pre-commit hook in the same way that we run on ci pipeline. I'll create separate issue for this to not forget it.

scripts/generate-react-exports.js
muratcorlu: We can inspire by Shoelace's react component wrapper generator: https://github.com/shoelace-style/shoelace/blob/next/scripts/make-react.js

baspinarenes: we didn't want to use custom-elements.json to be included in the build. then we can run `npm run analyze` before the generate step.

scripts/build.js
baspinarenes: okay. we will organize it as `baklava.ts` and `baklava-react.ts`.

docs/using-baklava-in-react.stories.mdx
baspinarenes: yes. it not needed to set the icon path when using it from cdn.

package.json
```diff
@@ -21,7 +21,7 @@
     "start": "npm run storybook:dev",
     "analyze": "cem analyze",
     "commit": "commit",
-    "build": "node scripts/build.js && npm run analyze",
+    "build": "npm run analyze && npm run generate-react-exports && node scripts/build.js && npm run analyze",
```
muratcorlu, line 24: We can generate react exports as last step here, right? 
```suggestion
    "build": "node scripts/build.js && npm run analyze && npm run generate-react-exports",
```

docs/using-baklava-in-react.stories.mdx
```diff
@@ -0,0 +1,91 @@
...
+Also, since React uses its own synthetic event system, it can't listen events that dispatches from Custom Elements.
+For this reasons, we used [@lit-labs/react](https://www.npmjs.com/package/@lit-labs/react) package to convert Custom Elements to React components.
+
+## Installation
+
+To add Baklava in your app, you can either import it via CDN or npm package.
+
+### Via CDN
+
+To be able to use Baklava via CDN, you should add our default.css and `react/index.js` at head tag in your `index.html` file.
```
muratcorlu, line 28: `react/index.js` part is outdated.

package.json
muratcorlu: Ah, no, build needs `src/baklava-react.ts` file. But I didn’t like to run analyze step two times here 😊

baspinarenes: yes, build needs `src/baklava-react.ts`. maybe we can generate `baklava-react.js` and add it to`dist` as manually after build.

muratcorlu: Source of problem here is that build script deletes `dist` folder before generating new build. I think it will be wise to extract this step from build script and clean dist folder in `npm run build` command. Like:

```
"build": "del-cli dist/ && npm run analyze && npm run generate-react-exports && node scripts/build.js"
```

You can add `del-cli` as dev dependency (we were already use `del` package).

scripts/generate-react-exports.js
```diff
@@ -0,0 +1,56 @@
+const fs = require('fs-extra');
+const prettier = require('prettier');
+
+function writeBaklavaReactFile(fileContentParts) {
+  let fileContentText = `
+    /* eslint-disable @typescript-eslint/ban-ts-comment */
+    // @ts-nocheck
+    import React from 'react';
```
muratcorlu, line 8: Currently "react" is bundled in our `baklava-react.js` file because of this import. Please add `external: ['react'],` to build js to exclude react from this file in the bundle.

scripts/generate-react-exports.js
```diff
@@ -0,0 +1,55 @@
...
+  const rawEventName = baklavaEventName.match(/(\w+)/g).at(-1);
+  return `on${rawEventName[0].toUpperCase()}${rawEventName.slice(1)}`;
+}
+
+const customElements = fs.readJSONSync(`${__dirname}/../dist/custom-elements.json`);
+const customElementsModules = customElements.modules;
+const baklavaReactFileParts = {
+  componentConverts: [],
+  exports: [],
+};
```
muratcorlu, line 31:
```suggestion
const baklavaReactFileParts = [];
```

scripts/generate-react-exports.js
```diff
@@ -0,0 +1,55 @@
...
+
+  baklavaReactFileParts.componentConverts.push(
+    `const ${componentName} = createComponent(
+      React,
+      '${fileName}',
+      customElements.get('${fileName}'),
+      ${JSON.stringify(eventNames)}
+    );`
+  );
+  baklavaReactFileParts.exports.push(componentName);
```
muratcorlu, line 52:
```suggestion
  baklavaReactFileParts.componentConverts.push(
    `export const ${componentName} = createComponent(
      React,
      '${fileName}',
      customElements.get('${fileName}'),
      ${JSON.stringify(eventNames)}
    );`
  );
```

scripts/generate-react-exports.js
```diff
@@ -0,0 +1,55 @@
...
+function writeBaklavaReactFile(fileContentParts) {
+  let fileContentText = `
+    /* eslint-disable @typescript-eslint/ban-ts-comment */
+    // @ts-nocheck
+    import React from 'react';
+    import { createComponent } from '@lit-labs/react';\n
+  `;
+
+  fileContentText += `${fileContentParts.componentConverts.join('\n')}\n\n`;
+  fileContentText += `export { ${fileContentParts.exports.join(', ')} }\n`;
```
muratcorlu, line 13:
```suggestion
    import { createComponent } from '@lit-labs/react';

    ${fileContentParts.join('\n')}
  `;
```

package.json
```diff
@@ -85,6 +88,7 @@
     "@web/test-runner-puppeteer": "^0.10.5",
     "@webcomponents/webcomponentsjs": "^2.5.0",
     "del": "^6.1.0",
```
muratcorlu, line 90: We don't need this in dependency list anymore (del-cli already has this). Please also import in build.js

muratcorlu (commented):
I have a suggestion to simplify auto generated file.

scripts/generate-react-exports.js
alibalbars: suggestions applied

scripts/generate-react-exports.js
alibalbars: suggestions applied

scripts/generate-react-exports.js
alibalbars: suggestions applied

muratcorlu (approved):

muratcorlu:
Looks great! Nice job! 🚀 @baspinarenes @alibalbars

Issue #151, opened 2022-07-04
Progress Component


batuhantozun:
## Design

[Figma Design Document
](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=151%3A2960)

## Implementation

General usage example:

```html
<bl-progress-indicator value="50"></bl-progress-indicator>
```

### Rules

* The progress indicator completely covers the area it is in.
* If the maximum value is not given, it is evaluated over 100.
* If you set the value greater than the maximum value, the value will be evaluated as the maximum value and the progress indicator will appear full.
* Progress indicator components should respect [accessibility attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role) inside.

### Usage Examples

The default max value is 100 so in the example below, the progress indicator will appear 25 percent full.

```html
<bl-progress-indicator value="25"></bl-progress-indicator>
```

For example, if you have a total of 9 tasks and completed 3 tasks you can pass parameters like this. The progress indicator will divide into 9 parts and 3 parts will be full.

```html
<bl-progress-indicator max="9" value="3"></bl-progress-indicator>
```

The default status is success if you want to show a fail progress indicator you should pass the `failed` property.

```html
<bl-progress-indicator value="75" failed></bl-progress-indicator>
```

If you want the progress indicator to appear full, you should set the value to the max value.

```html
<bl-progress-indicator max="12" value="12"></bl-progress-indicator>
<bl-progress-indicator value="100" failed></bl-progress-indicator>
```

#### Attributes

| Attribute | Description | Default Value |
| --------------- | --------------- | --------------- |
| size (`string`) | Size of progress component(small,medium,large) | medium |
| max (`string`) | Explains how many parts the specified progress indicator will be divided. It must have a value greater than 0 and be a valid floating point number. | 100 |
| value (`string`) | Specifies how much of the task has been completed. It must be a valid floating point number between 0 and max, or between 0 and 100 if max is omitted.  | - |
| failed (`boolean`) | Sets fail status | false |

DamlaDemir:
## Design

[Figma Design Document
](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=151%3A2960)

## Implementation

General usage example:

```html
<bl-progress-bar value='50'></bl-progress-bar>
```

### Rules

* The progress bar completely covers the area it is in.
[...]

PR #184, opened 2022-08-04
feat: progress-indicator component

DamlaDemir:
progress indicator component completed.

Issue https://github.com/Trendyol/baklava/issues/151


src/components/progress-indicator/bl-progress-indicator.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+import { Meta, Canvas, ArgsTable, Story, Preview, Source } from '@storybook/addon-docs';
+
+<Meta
+  title="Components/Progress Indicator"
+  component="bl-progress-indicator"
+  argTypes={{
+    size: {
+      control: {
+        type: 'select',
+        options: ['medium', 'large']
```
muratcorlu, line 14: small?

src/components/progress-indicator/bl-progress-indicator.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+    },
+  }}
+/>
+
+export const ProgressIndicatorTemplate = (args) => html`
+<bl-progress-indicator
+    size='${ifDefined(args.size)}'
+    max='${ifDefined(args.max)}'
+    value='${ifDefined(args.value)}'
+    failed='${ifDefined(args.failed)}'
```
muratcorlu, line 37:
```suggestion
    ?failed=${args.failed}
```

src/components/progress-indicator/bl-progress-indicator.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+</Canvas>
+
+## Progress Indicator Sizes
+
+We have 3 sizes of progress indicator: `large`, `medium`, `small`. Default size is `medium`.
+
+<Canvas>
+  <Story name="Progress Indicator Sizes">
+    {SizeTemplate.bind({})}
+  </Story>
```
muratcorlu, line 99: Add every size as a separate story. Then you'll not need to manually add margin for them and visual tests will be more clear.

src/components/progress-indicator/bl-progress-indicator.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+By default, the `max` is 100 and the progress indicator is evaluated over 100. So the `value` must be a valid floating point number between 0 and `max`, or between 0 and 100 if `max` is omitted.
+
+
+<Canvas>
+  <Story name="Basic Usage" args={{ value: '50' }}>
+    {ProgressIndicatorTemplate.bind({})}
+  </Story>
+</Canvas>
+
+You don't have to pass the `value` according to 100 percent. For example, if you have a total of 8 tasks and completed 5 tasks you can pass parameters like `max='8' value='5'`. The progress indicator will divide into 8 parts and 5 parts will be full.
```
muratcorlu, line 75: Please use double quotes for HTML attribute values (`max="8" value="5"`)

src/components/progress-indicator/bl-progress-indicator.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @property({ type: String })
+  size: ProgressIndicatorSize = 'medium';
+
+  @property({ type: String })
+  max = '100';
+
+  @property({ type: String })
+  value = '0';
```
muratcorlu, line 25: Use `number` type for those properties (as we did for `max` and `min` properties in `bl-input`)

src/components/progress-indicator/bl-progress-indicator.ts
```diff
@@ -0,0 +1,47 @@
...
+  @property({ type: Boolean })
+  failed = false;
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    setTimeout(() => {
+      this.style.setProperty('--bl-progress-indicator-max', this.max);
+      this.style.setProperty('--bl-progress-indicator-value', this.value);
+    });
```
muratcorlu, line 36: Those values should be set inside setters of max and value attributes. Because they can be changed over time.

muratcorlu: Can you try to read those values from attributes in CSS instead of setting in JS? https://developer.mozilla.org/en-US/docs/Web/CSS/attr If this works, it'll be easier, I think.

src/components/progress-indicator/bl-progress-indicator.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    el.setAttribute('failed', 'true');
+
+    //when
+    await elementUpdated(el);
+
+    //then
+    expect(el.size).to.eq('small');
+    expect(el.max).to.eq('5');
+    expect(el.value).to.eq('3');
+    expect(el.failed).to.eq(true);
```
muratcorlu, line 77: I think it'll be wise to check actual width of the progress bar (or at least related css variable) here. Because actually we are not updating progress bar width currently but this test passes.

src/components/progress-indicator/bl-progress-indicator.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+      type: 'string'
+    },
+    max: {
+      control: 'text',
+      type: 'string'
+    },
+    value: {
+      control: 'text',
+      type: 'string'
+    },
```
muratcorlu, line 25: Types should be number

muratcorlu:
I don't know why but controls for progress indicator canvas in Storybook doesn't work. Can you check?

image

muratcorlu:
Also there are some a11y remarks from Storybook.

image

src/components/progress-indicator/bl-progress-indicator.stories.mdx
DamlaDemir: What is the different between `ifDefined` and `?` expression?

src/components/progress-indicator/bl-progress-indicator.ts
DamlaDemir: I tried to use `attr` before but it doesn't work for me.

When I check in host:before i can see the values on the screen like this;
```
:host {
  --bl-progress-indicator-max: attr(max);
  --bl-progress-indicator-value: attr(value);
}

:host::before {
  content: var(--max) var(--value);
}
```

But when I tried accessing the `progress-indicator` class, it didn't work.

```
.progress-indicator::before {
  content: var(--value);
}
```

src/components/progress-indicator/bl-progress-indicator.ts
DamlaDemir: Actually, I used the number first but if I use the number i have to set property like this:

```
   this.style.setProperty('--bl-progress-indicator-max', this.max.toString());
   this.style.setProperty('--bl-progress-indicator-value', this.value.toString());
```
I didn't want to use `toString()` method and we are not calculating in js with these values. So I thought it can be` string`.

src/components/progress-indicator/bl-progress-indicator.stories.mdx
DamlaDemir: When I add stories for every size don't show the progress indicators. Because storybook adds div every story and this div width is 0 (I guess it works its child's width). Progress indicator width is 100% it works by the parent. So the component doesn't appear in the storybook. Maybe we can use different canvases.

src/components/progress-indicator/bl-progress-indicator.stories.mdx
muratcorlu: `?` is used for boolean attributes. If you don't use it result becomes `failed="true"` or `failed="false"`

src/components/progress-indicator/bl-progress-indicator.ts
muratcorlu: You could use `${this.max}`.

src/components/progress-indicator/bl-progress-indicator.ts
muratcorlu: You can define values on `.progress-indicator` element by reading them from aria attributes that you already write:

```css
.progress-indicator {
   --max: attr(aria-valuemax);
   --value: attr(aria-value);
  ...
}

.progress-indicator::before {
   width: calc(100% / var(--max) * var(--value));
   ...
}
```

With this, you don't need to listen and update css variables in JS anymore. Also setting private variable on `:host` element moved them to light dom (outside of the component). We already have kinda a problem about that and here we can start -maybe- a better approach to always set internal variables on wrapping element inside component.

src/components/progress-indicator/bl-progress-indicator.stories.mdx
muratcorlu: So this doesn't work?:

```jsx
<Canvas>
   <Story name="Progress Indicator Small" args={{ size: 'small' }}>
     {ProgressIndicatorTemplate.bind({})}
   </Story>
   <Story name="Progress Indicator Medium" args={{ size: 'medium' }}>
     {ProgressIndicatorTemplate.bind({})}
   </Story>
   <Story name="Progress Indicator Large" args={{ size: 'large' }}>
     {ProgressIndicatorTemplate.bind({})}
   </Story>
</Canvas>
```

src/components/progress-indicator/bl-progress-indicator.ts
muratcorlu: Apparently browser support for using `attr()` for the properties other than `content` is very bad: https://caniuse.com/css3-attr We can dismiss this for now.

But my first comment is still valid.

src/components/input/bl-input.stories.mdx
```diff
@@ -62,9 +62,6 @@ import {
     'label-fixed': {
       control: 'boolean',
     },
-    '': {
-      control: 'text',
-    },
```
muratcorlu, line 67: 👏🏻

muratcorlu (approved):

buseselvi:
Can we make the texts we show on the progress indicator in the storybook Sub Title 4 Regular style? 🙏🏻

leventozen (approved):

buseselvi (approved):

Issue #183, opened 2022-08-02
Visual regression tests for native element states in Storybook


muratcorlu:
This is a follow-up issue of #146 In that issue we could manage to run interaction tests for our tooltip component to get screenshots of the tooltip. But having same for input component was a bit more trickier. In this issue we'll target to write interaction tests for input focus, filled, valid, invalid states.

PR #187, opened 2022-08-04
fix(button): outline hover state fixed

muratcorlu:
We noticed that hover style of button's `outline` variant was not correct. This fixes that problem and also adds visual tests for hover states of the buttons. So hopefully we'll not miss a problem like this again.

This is also related to #183. I added a storybook plugin to simulate css :hover effect in Storybook stories. I'll also continue on this for input component (it has a bit trickier situation) in a separate work.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -23,7 +23,6 @@ module.exports = {
     '@storybook/addon-essentials',
     '@storybook/addon-a11y',
     '@storybook/addon-interactions',
-    'storybook-addon-pseudo-states',
   ],
   features: {
     postcss: false,
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -23,6 +23,7 @@ module.exports = {
     '@storybook/addon-essentials',
     '@storybook/addon-a11y',
     '@storybook/addon-interactions',
+    'storybook-addon-pseudo-states',
   ],
   features: {
     postcss: false,
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,7 +60,6 @@
         "rimraf": "^3.0.2",
         "rollup-plugin-lit-css": "^4.0.0",
         "storybook": "6.5.9",
-        "storybook-addon-pseudo-states": "^1.15.1",
         "stylelint": "^14.6.1",
         "stylelint-config-prettier": "^9.0.3",
         "stylelint-config-standard": "^25.0.0",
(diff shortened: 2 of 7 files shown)


muratcorlu:
Somehow snapshots are not correct again on Chromatic. :( I'm following the issue here: https://github.com/chromaui/chromatic-cli/issues/622

buseselvi (approved):

leventozen (approved):

leventozen:
When I tried to run on my local, it won't work for storybook for me. However, when I checked [chromatic](https://www.chromatic.com/build?appId=6267ebb4dd54f5004a65f14a&number=215) build, seems like it worked 🤔 

I faced this particular behaviour: _storybook-addon-pseudo-states_ is adding pseudo class at higher level component and it doesn't affect inner elements at all in Storybook for me. Right now, I found a way to fix this problem by giving id to the element and add class for this particular id like this without changing the structure. 
```js
export const states = {
  pseudo: {
    hover: ["#hover-state"]
  }
}
```
```js
<Canvas isColumn>
  <Story name="Contained Buttons Hover" args={{ id: 'hover-state', content: 'Save', icon: 'info' }} parameters={hover}>
    {Template.bind({})}
  </Story>
  <Story name="Outline Buttons Hover" args={{ id: 'hover-state', content: 'Save', outline: true, icon: 'info' }} parameters={hover}>
    {Template.bind({})}
  </Story>
    <Story name="Text Buttons Hover" args={{ id: 'hover-state', content: 'Save', text: true, icon: 'info' }} parameters={hover}>
    {Template.bind({})}
  </Story>
</Canvas>
```

```js
export const SingleButtonTemplate = (args) => html`<bl-button
    id=${ifDefined(args.id)}
    ?primary=${args.primary}
    ?secondary=${args.secondary}
    ?success=${args.success}
    ?danger=${args.danger}
    ?outline=${args.outline}
    ?disabled=${args.disabled}
    ?text=${args.text}
    href=${ifDefined(args.href)}
    target=${ifDefined(args.target)}
    size=${ifDefined(args.size)}
    icon="${ifDefined(args.icon)}"
    label="${ifDefined(args.label)}"
    style=${ifDefined(args.styles ? styleMap(args.styles) : undefined)}
      >${unsafeHTML(args.content)}</bl-button>`
```


However, I decided to work at this issue at https://github.com/Trendyol/baklava/issues/183.
So, I will extract the pseudo-states part from this PR and merge the outline hover state fix.

PR #188, opened 2022-08-05
fix(badge): css variables used as privately

muratcorlu:
As we discussed in #164, it was not possible to set badge CSS variables from upper levels of dom tree because default values set in `:host` selector. This was causing difficulties to style badge in `tab` component, because we use badge inside tab component's shadow dom and it's not possible to query badge inside tab in CSS. This makes CSS variables of badge component like a theme variables, so user can set those variables in any scope (badge component level, parent component level or document level).

### DIFF
--- a/src/components/badge/bl-badge.css
+++ b/src/components/badge/bl-badge.css
@@ -30,12 +30,12 @@
   height: var(--height);
 }
 
-:host([size=small]) .badge {
+:host([size='small']) .badge {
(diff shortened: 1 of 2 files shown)


muratcorlu:
@buseselvi I added 2 more stories to test small badge with icon (we hide icon in this case) and large badge with icon. Can you please check them on Chromatic?

DamlaDemir (approved):

buseselvi (approved):

PR #191, opened 2022-08-08
fix(button)!: button attributes refactored

muratcorlu:
**This is a BREAKING CHANGE**

Currently we use boolean attributes for button variants (`primary`, `secondary`, `success`, `danger`) and types (`contained`, `outline`, `text`). In the beginning we thought this is easier to write but currently I see that this is conflicting with the approach how we use boolean attributes in other components and how we really should use them. Boolean attributes should not rely on other attributes. But now we are technically allowed to write `<bl-button primary secondary>` and in this case result relies on our CSS specifity. And this type of writing does give the feeling that it's possible.

I think we need to fix this as soon as possible to not make confusion for new comers.

This PR converts boolean variants to a more standart `variant` attribute with a default `primary` value again. I picked `fill` attribute value for button types (we use type in storybook) because probably `type` attribute will be needed in near future for using a button as a submitter of a form. Since HTML `button` has `type` I thought we should not use `type` for setting button as `outline` or `text`. Opinions are very welcome.

### DIFF
--- a/src/components/button/bl-button.css
+++ b/src/components/button/bl-button.css
@@ -76,28 +76,28 @@
   --bl-button-margin-icon: 0;
 }
 
-:host([variant=secondary]) {
+:host([variant='secondary']) {
   --bl-button-main-color: var(--bl-color-secondary);
   --bl-button-main-hover-color: var(--bl-color-secondary-hover);
 }
 
-:host([variant=success]) {
+:host([variant='success']) {
   --bl-button-main-color: var(--bl-color-success);
   --bl-button-main-hover-color: var(--bl-color-success-hover);
 }
 
-:host([variant=danger]) {
+:host([variant='danger']) {
   --bl-button-main-color: var(--bl-color-danger);
   --bl-button-main-hover-color: var(--bl-color-danger-hover);
 }
 
-:host([fill=text]) {
+:host([fill='text']) {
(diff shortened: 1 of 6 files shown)


DamlaDemir (approved):

DamlaDemir:
Good job!

leventozen:
Good work! However, I believe using `button-type` is more clear than `fill`. Because in our design system, we are naming `contained`, `outline` and `text` as button types.

leventozen:
After discuss with @DamlaDemir, we think that `type` and `button-type` may confuse people when they need to use them at the same time. We decided to go with `kind` because it is a synonym of `type`.

buseselvi:
can we fix the buttons on the tab component to secondary text button - icon only variant like before?

leventozen:
@buseselvi we fixed it, can you check it again?

buseselvi:
@leventozen buttons look ok 👌  but this time in the hover state tooltip seems to slide down 
image

leventozen:
> @leventozen buttons look ok 👌 but this time in the hover state tooltip seems to slide down image

@buseselvi, this changes are not related to this bug, so we will create another issue. thank you for your attention 🤖

buseselvi (approved):

leventozen (approved):

PR #193, opened 2022-08-09
docs: add Styling Components section to react doc

alibalbars:
I think it would be useful for React users to see directly from the doc how to style components.

### DIFF
--- a/docs/using-baklava-in-react.stories.mdx
+++ b/docs/using-baklava-in-react.stories.mdx
@@ -125,8 +125,8 @@ import { BlButton } from "@trendyol/baklava/dist/baklava-react";
 
 function MyComponent() {
   const buttonStyle = {
-    "--bl-button-main-color": "rebeccapurple",
-    "--bl-button-main-hover-color": "purple",
+    "--bl-color-primary": "purple",
+    "--bl-color-primary-hover": "rebeccapurple",
   }
 
   return (
@@ -145,11 +145,11 @@ import styled from "styled-components"
 
 const Wrapper = styled.div`
    .button-class {
-     --bl-button-main-color: red;
+     --bl-color-primary: purple;
    }
-
+   
    .tooltip-class {
-     --bl-tooltip-top: 100px !important;
+     --bl-color-primary-hover: rebeccapurple;
    }
 `;
 
--- a/docs/using-baklava-in-react.stories.mdx
+++ b/docs/using-baklava-in-react.stories.mdx
@@ -147,9 +147,9 @@ const Wrapper = styled.div`
    .button-class {
      --bl-color-primary: purple;
    }
-   
+
    .tooltip-class {
-     --bl-color-primary-hover: rebeccapurple;
+     --bl-tooltip-position: fixed;
    }
 `;
 
--- a/docs/using-baklava-in-react.stories.mdx
+++ b/docs/using-baklava-in-react.stories.mdx
@@ -114,3 +114,55 @@ function MyComponent() {
 
 export default MyComponent;
 ```
+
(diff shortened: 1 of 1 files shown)


docs/using-baklava-in-react.stories.mdx
```diff
@@ -114,3 +114,55 @@ function MyComponent() {
...
+#### Inline CSS
+
+```jsx
+import { BlButton } from "@trendyol/baklava/dist/baklava-react";
+
+function MyComponent() {
+  const buttonStyle = {
+    "--bl-button-main-color": "rebeccapurple",
+    "--bl-button-main-hover-color": "purple",
+  }
```
muratcorlu, line 130: This is not intended behavior. We'll move those "private" variables somewhere else. In theory it's not possible to change a color of a button variant without setting related theme color. But since you can set theme variables for a specific scope in document, you can set it like that:

```suggestion
  const buttonStyle = {
    "--bl-color-primary": "rebeccapurple",
    "--bl-color-hover": "purple",
  }
```

docs/using-baklava-in-react.stories.mdx
```diff
@@ -114,3 +114,55 @@ function MyComponent() {
...
+import { BlTooltip, BlButton } from "@trendyol/baklava/dist/baklava-react";
+import styled from "styled-components"
+
+const Wrapper = styled.div`
+   .button-class {
+     --bl-button-main-color: red;
+   }
+
+   .tooltip-class {
+     --bl-tooltip-top: 100px !important;
```
muratcorlu, line 152: This is not a good example. You can not (should not) set tooltip position like that (at least for now).

docs/using-baklava-in-react.stories.mdx
```diff
@@ -114,3 +114,55 @@ function MyComponent() {
...
+
+#### Styled Components
+
+```jsx
+import { BlTooltip, BlButton } from "@trendyol/baklava/dist/baklava-react";
+import styled from "styled-components"
+
+const Wrapper = styled.div`
+   .button-class {
+     --bl-button-main-color: red;
```
muratcorlu, line 148: Same applies here as above.

muratcorlu (changes requested):

docs/using-baklava-in-react.stories.mdx
alibalbars: thanks for info. fixed

docs/using-baklava-in-react.stories.mdx
alibalbars: fixed, changed with color example

docs/using-baklava-in-react.stories.mdx
alibalbars: fixed

docs/using-baklava-in-react.stories.mdx
```diff
@@ -114,3 +114,55 @@ function MyComponent() {
...
+import { BlTooltip, BlButton } from "@trendyol/baklava/dist/baklava-react";
+import styled from "styled-components"
+
+const Wrapper = styled.div`
+   .button-class {
+     --bl-color-primary: purple;
+   }
+   
+   .tooltip-class {
+     --bl-color-primary-hover: rebeccapurple;
```
muratcorlu, line 152: Tooltip doesn't use this variable. You can set `--bl-tooltip-position: fixed;` as an example.

alibalbars: fixed. this example worked so i thought it was available in the tooltip

muratcorlu (approved):

PR #195, opened 2022-08-09
chore: pipeline and commit hook linters are synced

### LINKED ISSUE #182: Linter check improvements
- [ ] Pre-commit hook linter doesn't work in the same way of `npm run lint` we commit our work and linter step in CI/CD pipeline fails very often. We need to catch lint errors before committing them.
- [ ] `npm run lint` report is not very informative. We need to check options to see more detailed error (or warning) messages.

muratcorlu:
There were some difference between pre-commit hook (`lint-staged`) and pipeline `npm run lint` about linting configurations. This PR syncs them.

Now `npm run lint` and `lint-staged` checks same rules. `npm run format` runs prettier to automatically fix some linting errors. If some checks are failed in `lint-staged` command, you can try this auto formatting command.

PS: I couldn't test these changes very well. And since I'll be on vacation starting from tomorrow, I wanted to share current version as a PR. Please be more critical on changes and feel free to add changes or reject it if you see any problem.

Fixes #182


muratcorlu:
I removed mdx linter from this PR because couldn't manage to run it. Updated linters document. Also did small arrangement on docs by moving icon tutorial to a more related place. @leventozen can you review this version?

.lintstagedrc
```diff
@@ -1,9 +1,11 @@
 {
-  "*.{ts,tsx,js}": [
-    "eslint --rule 'no-console: error' --fix",
-    "prettier --write src/"
+  "*.ts": [
+    "tsc --noEmit"
```
leventozen, line 3: It won't work since when lint-staged is called via husky hooks, `tsc` will omit `tsconfig.json`. 
I believe, we have two options here

- We can go with `bash -c tsc --noEmit` to make it run with the context.
- Or we can use function syntax like this: `() => "tsc --noEmit"`. To make it run, we should change the config file extension to js obviously.

I would prefer the second option here to support cross-platform development. 🤔

leventozen: I didn't check yet but there could be another option. We can specify the config file by using `-p` parameter.

src/components/input/bl-input.ts
```diff
@@ -179,14 +179,16 @@ export default class BlInput extends LitElement {
       : '';
     const label = this.label ? html`<label>${this.label}</label>` : '';
 
+    const classes = {
```
leventozen, line 182: We can write it like this to make this cleaner. Also, for consistentcy with the button

```js
const classes = classMap({
      'dirty': this.dirty,
      'has-icon': this.icon || (this.dirty && this._invalidState),
      'has-value': this.hasValue,
});
```

leventozen (approved):

PR #201, opened 2022-08-17
feat: checkbox component

alibalbars:
this pr is for new checkbox component. 

- changed indeterminate logic ( single checkbox can be indeterminate )
- removed checkbox-group element
- added storybook file
- added test file

Problem: 
- there is a css bug when changing state of a single checkbox (you can see it in storybook)
  Adding `:host { display: flex }` to `bl-checkbox.css` fixed it but it makes the element like `display: block`


src/components/checkbox-group/bl-checkbox-group.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * @tag bl-checkbox-group
+ * @summary Baklava Checkbox Group component
+ */
+
+@customElement('bl-checkbox-group')
+export default class BlCheckboxGroup extends LitElement {
+  @query('#parentCheckbox')
+  parentCheckbox: any;
```
DamlaDemir, line 13: Please don't use `any` here.

src/components/checkbox-group/bl-checkbox-group.ts
```diff
@@ -0,0 +1,82 @@
...
+    } else {
+      this._slottedChildren.forEach(item => {
+        item.checked = false;
+      });
+    }
+  }
+
+  render(): TemplateResult {
+    return html`
+        <div style="display: flex; flex-direction: column;">
```
DamlaDemir, line 66: We don't use inline style please use CSS class or tag selectors.

src/components/checkbox-group/checkbox/bl-checkbox.css
```diff
@@ -0,0 +1,67 @@
+:host {
+    display: inline-block;
+
+    --bg-color: var(--bl-checkbox-bg-color, var(--bl-color-primary));
```
DamlaDemir, line 4: As we discussed in https://github.com/Trendyol/baklava/discussions/164 we don't define CSS variable in the host anymore. You can define in box class scope.

src/components/checkbox-group/checkbox/bl-checkbox.css
```diff
@@ -0,0 +1,67 @@
...
+    --bg-color-disabled: var(--bl-color-tertiary, var(--bl-color-disabled));
```
DamlaDemir, line 5: Actually, all those variables don't change by any prop or attributes, we can set them directly. We don't have to define another variable. You can set it like this:

```
    width: var(---bl-size-m);
    height: var(--bl-size-m);
```

src/components/checkbox-group/checkbox/bl-checkbox.css
```diff
@@ -0,0 +1,67 @@
...
+    height: var(--height);
+    border: 1px solid var(--border-color);
+    border-radius: var(--box-radius);
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+.box .icon {
+    font-size: 8px;
```
DamlaDemir, line 29: please use our sizing variable.

src/components/checkbox-group/checkbox/bl-checkbox.ts
```diff
@@ -0,0 +1,96 @@
...
+
+@customElement('bl-checkbox')
+export default class BlCheckbox extends LitElement {
+  static get styles(): CSSResultGroup {
+    return [style];
+  }
+  private checkboxGroup: BlCheckboxGroup | null;
+
+  @query('#checkbox')
+  checkbox: any;
```
DamlaDemir, line 20: We don't use `any` you can use more specific `HTMLInputElement`.

src/components/checkbox-group/checkbox/bl-checkbox.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  @query('#label')
```
DamlaDemir, line 22: Same as above don't use `any` you can use more specific `HTMLLabelElement`.

src/components/checkbox-group/checkbox/bl-checkbox.ts
```diff
@@ -0,0 +1,96 @@
...
+  labelEl: any;
+
+  @property({ type: String, reflect: true })
+  label = '';
```
DamlaDemir, line 26: The label should be optional because we use the checkbox without a `label` in the data table component.

src/components/checkbox-group/checkbox/bl-checkbox.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  @property({ type: Boolean, reflect: true })
+  checked = false;
+
+  @property({ type: Boolean, reflect: true })
+  disabled = false;
+
+  @property({ type: Boolean, reflect: true })
+  indeterminate = false;
```
DamlaDemir, line 35: The `indeterminate` feature should be thought about again. Because when I discuss with @buseselvi she said only the first checkbox in the group checkbox can be `indeterminate`. But when I try I can set any checkbox as `indeterminate`. We shouldn't set as `indeterminate` any checkbox.

src/components/checkbox-group/checkbox/bl-checkbox.ts
```diff
@@ -0,0 +1,96 @@
...
+    checked: 'check',
+    indeterminate: 'minus',
+  };
+
+  icons = {
+    checked: html`<bl-icon class="icon" name="check"></bl-icon>`,
+    indeterminate: html`<bl-icon class="icon" name="minus"></bl-icon>`,
+  };
+
+  handleChange(event: any) {
```
DamlaDemir, line 49: Same as above don't use `any` you can fix like this:

```
  changeHandler(event: Event) {
      const checked = (event.target as HTMLInputElement).checked;
      this.checked = checked;
      this.onChange(checked);
   }
```

src/components/checkbox-group/checkbox/bl-checkbox.ts
```diff
@@ -0,0 +1,96 @@
...
+    this.checked = event.target.checked;
+    this.onChange(event.target.checked);
+
+    if (this.indeterminate) {
```
DamlaDemir, line 53: Is this `if condition` necessary? I think we can set directly  `this.indeterminate = false;`

src/components/checkbox-group/checkbox/bl-checkbox.ts
```diff
@@ -0,0 +1,96 @@
...
+      this.indeterminate = false;
+    }
+  }
+
+  updated(changedProperties: Map<string, unknown>) {
+    if (this.indeterminate) {
```
DamlaDemir, line 59: I couldn't catch why we set `this.checkbox.indeterminate`. Where do we use this?

src/components/checkbox-group/checkbox/bl-checkbox.ts
```diff
@@ -0,0 +1,96 @@
...
+    return html`
+    <label id="label" class="label">
+      <input id="checkbox" type="checkbox" name="checkbox" ?checked=${this.checked} ?disabled=${
+      this.disabled
+    } @change=${this.handleChange} />
+      <div class="box">
+        ${this.indeterminate && !this.checked ? this.icons.indeterminate : null}
+        ${this.checked ? this.icons.checked : null}
+      </div>
+      <span class="text">${this.label}</span>
```
DamlaDemir, line 86: When set as optional the label we should use `ifDefined` here.

src/components/checkbox-group/bl-checkbox-group.ts
alibalbars: fixed

src/components/checkbox-group/bl-checkbox-group.ts
alibalbars: fixed

src/components/checkbox-group/checkbox/bl-checkbox.css
alibalbars: fixed

src/components/checkbox-group/checkbox/bl-checkbox.css
alibalbars: fixed

src/components/checkbox-group/checkbox/bl-checkbox.ts
alibalbars: I changed the logic it as you said. now only the first checkbox in the group can have indeterminate state. but to do this, i add a new property called `allowIndeterminate` to `bl-checkbox`. unfortunately it shows up in docs. I couldn't do it any other way

src/components/checkbox-group/checkbox/bl-checkbox.ts
alibalbars: fixed

src/components/checkbox-group/checkbox/bl-checkbox.ts
alibalbars: fixed

src/components/checkbox-group/checkbox/bl-checkbox.ts
alibalbars: If I understood your question correctly, indeterminate state cannot be set using an HTML attribute.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes

src/components/checkbox-group/checkbox/bl-checkbox.ts
alibalbars: fixed

src/components/checkbox-group/checkbox/bl-checkbox.ts
alibalbars: fixed

src/components/checkbox-group/checkbox/bl-checkbox.ts
alibalbars: fixed

src/components/checkbox-group/checkbox/bl-checkbox.ts
alibalbars: fixed

src/components/checkbox-group/checkbox/bl-checkbox.ts
muratcorlu: Why we need a logic like "only first checkbox can be `indeterminate`"? I never saw a restriction like this before. Generally indeterminate checkboxes are used in tree-like selections. Parent of a checkbox group can be indeterminate state regarding to selection of sub checboxes. But it doesn't need to be first checbox in the parent group.

src/components/checkbox-group/checkbox/bl-checkbox.ts
```diff
@@ -0,0 +1,105 @@
...
+        @change=${this.handleChange} />
+      <div class="box">
+        ${this.checked ? this.icons.checked : null}
+        ${
+          this.indeterminate && this.indeterminateAllowed && !this.checked
+            ? this.icons.indeterminate
+            : null
+        }
+      </div>
+      ${ifDefined(this.label) && html`<span class="text">${this.label}</span>`}
```
muratcorlu, line 95: I think it's better to use slot for label, instead of attribute.

muratcorlu: Since we are having this discussion on nearly every component, I'm doing some investigation on this topic now. I'll share outcome soon.

src/components/checkbox-group/checkbox/bl-checkbox.ts
DamlaDemir: Actually, I wanted to say only the parent checkbox of the group can be `indeterminate`.

src/components/checkbox-group/checkbox/bl-checkbox.ts
alibalbars: okey, i will change it

buseselvi:
text is sliding down when we check the checkbox in storybook, can we fix it?

alibalbars:
> text is sliding down when we check the checkbox in storybook, can we fix it? Also, can we make these two aligned? 🙏🏻 image

i tried but couldn't fix it. i need some help here 🙏

umutbozdag:
> > text is sliding down when we check the checkbox in storybook, can we fix it? Also, can we make these two aligned? 🙏🏻 image
> 
> i tried but couldn't fix it. i need some help here 🙏
Probably you will get the result you want when you change `label`s display to `flex` instead of `inline-flex`:
image
image
image

src/components/checkbox/bl-checkbox.ts
```diff
@@ -0,0 +1,87 @@
...
+          name="checkbox"
+          ?checked=${this.checked}
+          ?disabled=${this.disabled}
+          @change=${this.handleChange}
+        />
+        <div class="box">
+          ${this.checked ? this.icons.checked : null}
+          ${this.indeterminate && !this.checked ? this.icons.indeterminate : null}
+        </div>
+        <span class="text"><slot></slot></span>
```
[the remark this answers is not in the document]
umutbozdag: Are you sure that wrapping with `label` is the correct way for it?

umutbozdag:
> > > > text is sliding down when we check the checkbox in storybook, can we fix it? Also, can we make these two aligned? 🙏🏻 image
> > > 
> > > 
> > > i tried but couldn't fix it. i need some help here 🙏
> > 
> > 
> > Probably you will get the result you want when you change `label`s display to `flex` instead of `inline-flex`: image image image
> 
> yes it solves the jumping issue, but i makes the checkbox element block :(
> 
> Screen Shot 2022-08-29 at 18 49 07

I think you could use `flex` in the wrapper `div` to make it look like side by side. I think side by side feature is only for showcase and I am not sure if it's really need to be an inline component. Maybe you guys would want to develop something like `checkbox-group` component to wrap checkboxes in the future. @buseselvi @alibalbars

muratcorlu:
@alibalbars @umutbozdag I made some changes. I set host element `inline-block` and setting `vertical-align` helped fixing the issue. Please review current version.

alibalbars:
> @alibalbars @umutbozdag I made some changes. I set host element `inline-block` and setting `vertical-align` helped fixing the issue. Please review current version.

i came across with some bugs. first one;

https://user-images.githubusercontent.com/43646356/187701076-0a2edb8c-9372-4e4a-9770-bc8a5ba24c4b.mov

Infos: 
- react project,
- buttons connected with checkbox

Bug Senario:
1- click checkbox
2- click checked button twice
3- first click to checkbox not working, but second is working

for debug i printed actual input elements checked value and lit elements checked property. This are different in this case. i solved this by setting actual input element's checked value in `updated` method.

alibalbars:
> @alibalbars @umutbozdag I made some changes. I set host element `inline-block` and setting `vertical-align` helped fixing the issue. Please review current version.

second strange case;


`<BlCheckbox checked indeterminate>test</BlCheckbox>` show this result, 
weird part is checkbox showing indeterminate icon but label is orange. i know that the label should be orange only when it is checked. 


image

muratcorlu:
> > @alibalbars @umutbozdag I made some changes. I set host element `inline-block` and setting `vertical-align` helped fixing the issue. Please review current version.
> 
> second strange case;
> 
> `<BlCheckbox checked indeterminate>test</BlCheckbox>` show this result, weird part is checkbox showing indeterminate icon but label is orange. i know that the label should be orange only when it is checked.
> 
> image

I tried to copy behaviour of native checkbox about indeterminate state but now I realized I misunderstood the native behavior. I think `indeterminate` state should to override `checked` value to `false`. But `checked` state should also override `inderminate` to `false`. But what should happen if user set `<bl-checkbox checked inderminate>` ? I’m not sure.

muratcorlu:
@buseselvi What will be the color of the checkbox's label when it's in `disabled` state? And will it be in same color if its `disabled` and `checked`?

buseselvi:
> @buseselvi What will be the color of the checkbox's label when it's in `disabled` state? And will it be in same color if its `disabled` and `checked`? Can you provide this information also in the design document?

Color of the checkbox's labels should be content passive (#AFBBCA). I updated the design document accordingly 🙏🏻

src/components/checkbox/bl-checkbox.ts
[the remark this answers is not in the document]
muratcorlu: Yes. 

> Alternatively, you can nest the <input> directly inside the <label>, in which case the for and id attributes are not needed because the association is implicit:
> 
> ```html
> <label>Do you like peas?
>  <input type="checkbox" name="peas">
> </label>
> ```
>
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

buseselvi (approved):

stovein:
With last commits, it should cover all necessary states. Can you please check it?

src/components/checkbox/bl-checkbox.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (this.indeterminate) icon = 'minus';
+
+    return html`
+      <label>
+        <input
+          type="checkbox"
+          name="checkbox"
+          .checked=${live(this.checked)}
+          ?disabled=${this.disabled}
+          .indeterminate=${this.indeterminate}
```
muratcorlu, line 68: Do we also need `live` here?

muratcorlu (approved):
I think this is good enough to continue with.

One remark: Using this as an input field will still need some extra work. Validation, FormData integration and ElementInternals integration are missing but I think it's ok to merge this without those and work on a separate issue for them.

src/components/checkbox/bl-checkbox.ts
muratcorlu: I meant also having live for indeterminate state like `.indeterminate=${live(this.indeterminate)}`

muratcorlu: Let's continue with this.

PR #209, opened 2022-08-24
feat: minify css

### LINKED ISSUE #175: Minify component CSS in ESM bundles
Component styles in ESM bundles are not minified. Example: https://cdn.jsdelivr.net/npm/@trendyol/baklava@2.0.0-beta.10/dist/chunk-CGZOADFQ.js

Minifying those styles can help making prod bundle a bit lighter.

Caution: Source maps needs to be considered.

leventozen:
There is a builtin option in our css plugin named [uglifycss](https://www.npmjs.com/package/uglifycss). So, we can just add `uglify: true` option to minify our css. Output seems good and I didn't want to add another step/plugin to minify the css. I believe, this'll work for us.

Closes #175

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -39,6 +39,7 @@ const args = parseArgs(process.argv.slice(2), {
       external: ['react'],
       plugins: [
         litCssPlugin({
+          uglify: true,
           filter: /components\/.*\.css$/,
         }),
       ],


muratcorlu (approved):
Easy peasy! 😃

muratcorlu:
@leventozen Maybe this can be considered as a feature and change commit prefix (PR title) as `feat:`. Because `perf` doesn't trigger a new release but optimization on build bundle size actually has a good value for the consumers of Baklava.

Issue #206, opened 2022-08-23
Visual regression test for button component states


batuhantozun:
We need to create visual regression test for button component's hover state.

Issue #203, opened 2022-08-19
Hot reload doesn't watch component files


bbuergel:
**Describe the bug**
When running storybook locally, only story .mdx files are watched in Webpack hot reloads. Modifying component files require running the build command in order for Storybook to reload the changes.

**To Reproduce**

1. Clone/fork repo
2. npm install
3. npm run build
4. npm run start
5. Make edits to component .ts files and save

**Expected behavior**
Expecting to see Webpack hot reload refresh changes to component files, not just .mdx stories

**Desktop**
 - OS: MacOS 12.5.1
 - Chrome

muratcorlu:
Thanks for reporting this. I acknowledge the issue.

Storybook only listens to the build folder and story files. We use `esbuild` for building components, and Storybook uses webpack to serve its own UI together with our build output. We didn't want to use Storybook build flow since it caused some issues with styles and assets.

We'll have a look for the most efficient way of running Storybook in watch mode with our library build.

muratcorlu:
If you run `npm run serve` it basically runs a build in watch mode and serve it from `8000` port with the folder `playground` in the root. We are generally just copy `template.html` in playground folder with a different name like `my-component.html` and do our testing on that file.

Development directly on storybook was our first idea but because its build process is very heavy and complicated, we are currently using this approach. When basic stuff starts working, we start working on Storybook. And if you run storybook with `npm run storybook:dev` if actually watch `dist` folder too. So while storybook open, in another terminal window you can make a new build with `npm run build` storybook should refresh with new files.

Pickra:
Thanks @muratcorlu .

I saw that, I get both approaches; playground and `npm run build`. I find more value w/Storybook for local dev. It's better DX + I deploy the static build for my users, so it helps to be in the same environment. I'm experimenting w/[vite](https://vitejs.dev/). So far: 
- the HMR works great w/storybook
- Storybook static build is fine 
- lib package build looks good too

muratcorlu:
Good feedback, very appreciated! We also want to put Storybook at the center of our development process. It already covers our UI and behavior tests with the help of [Chromatic](https://chromatic.com). And we'll definitely work on improving the DX with Storybook.

I'll have a look this week at this topic.

Pickra:
I got Storybook to work w/vite HMR very easily @muratcorlu . I still haven't totally settled on vite but it's been great so far. It also means u can probably use most of, if not all, your esbuild + rollup biz.

Pickra:
[...]

PR #212, opened 2022-08-25
chore: add visual regression test for button component states

leventozen:
At first glance, we know we can add corresponding classes for every component state but we don't want to hardcode for every state. After that, we tried to use [storybook-addon-pseudo-states](https://github.com/chromaui/storybook-addon-pseudo-states) which was [suggested](https://github.com/chromaui/chromatic-cli/issues/633) by the chromatic team. However, we found out that it is not mature for lit. [Here](https://github.com/chromaui/storybook-addon-pseudo-states/issues/38) you can see our issue in this plugin and find detailed information.

Right now, we add corresponding classes for every component state but **automatically in build time**. So, we don't need to write classes for every component. Also, we have special classes `__ONLY_FOR_STORYBOOK_DEMONSTRATION_HOVER__`  to avoid confusion in the documentation.

Closes https://github.com/Trendyol/baklava/issues/206

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -10,7 +10,22 @@ const args = parseArgs(process.argv.slice(2), {
 (async () => {
   const { globby } = await import('globby');
   const destinationPath = 'dist';
-  const isProduction = process.env.NODE_ENV === 'production';
+  const isRelease = process.env.RELEASE || false;
+
+  const cssPluginOptions = {
+    uglify: true,
+    filter: /components\/.*\.css$/
+  };
+  
+  if (!isRelease) {
+    cssPluginOptions.transform = (content) => content.replace(/.*:hover[^{]*/g, matched => {
+      // Replace :hover with special class. (There will be additional classes for focus, etc. Should be implemented in here.)
+      const replacedWithNewClass = matched.replace(/:hover/, '.__ONLY_FOR_STORYBOOK_DEMONSTRATION_HOVER__')
+      // Concat strings
+      return matched.concat(', ', replacedWithNewClass)
+    })
+  }
+  
 
   try {
     const buildOptions = {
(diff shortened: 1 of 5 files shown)


buseselvi (approved):

scripts/build.js
```diff
@@ -10,6 +10,7 @@ const args = parseArgs(process.argv.slice(2), {
 (async () => {
   const { globby } = await import('globby');
   const destinationPath = 'dist';
+  const isProduction = process.env.NODE_ENV === 'production';
```
muratcorlu, line 13: I think we should do this "hack" only for storybook builds.

scripts/build.js
```diff
@@ -41,6 +42,13 @@ const args = parseArgs(process.argv.slice(2), {
         litCssPlugin({
           uglify: true,
           filter: /components\/.*\.css$/,
+          // Match line containing ':hover'
+          transform: (content) => isProduction ? content : content.replace(/.*:hover[^{]*/g, matched => {
```
muratcorlu, line 46: I would not do that check in every cycle. Instead something like:

```js
const cssPluginOptions = {
  uglify: true,
  filter: /components\/.*\.css$/
};

if (isStorybook) {
   cssOptions.transform = (content) => content.replace( .... 
}

...

litCssPlugin(cssPluginOptions),
```

leventozen: Good catch. 👍

scripts/build.js
leventozen: I think it should work for every build except releases. If I add a condition for only storybook builds, it will not work well in the development environment. Any advice for this?

src/components/button/bl-button.stories.mdx
```diff
@@ -209,6 +204,41 @@ We have 2 types of disabled buttons: **Contained** and **Text**. Disable version
...
+    variant=${ifDefined(args.variant)}
+    kind=${ifDefined(args.kind)}
+    href=${ifDefined(args.href)}
+    target=${ifDefined(args.target)}
+    size=${ifDefined(args.size)}
+    icon="${ifDefined(args.icon)}"
+    label="${ifDefined(args.label)}"
+    ?disabled=${args.disabled}
+    style=${ifDefined(args.styles ? styleMap(args.styles) : undefined)}
+      >${unsafeHTML(args.content)}</bl-button>`
```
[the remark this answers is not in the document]
leventozen: To be able to do this I should implement a parameter in the `SingleButtonTemplate` which I don't want to do at first. However, `class=${ifDefined(args.class)}` may have been considered because it won't affect other stories thanks to `ifDefined`.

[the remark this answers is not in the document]
muratcorlu: I thought same. I think it's better now.

scripts/build.js
muratcorlu: In development environment (like in playground, or linking to another project to test etc) we should never need that to use this class name. Class name express itself: `__ONLY_FOR_STORYBOOK_DEMONSTRATION_HOVER__` 😊

muratcorlu: There are some other works on storybook flow (#203). Let's fix this detail there. I think we can go with this.

muratcorlu (approved):

leventozen:
We decided to move with this solution with @muratcorlu. However, we will iterate this solution with new features that we want to add to the storybook. 
Here you can find detailed information about it.
https://github.com/Trendyol/baklava/issues/203

PR #213, opened 2022-08-25
chore: describe border radius variables

### LINKED ISSUE #185: Describe Border Radius Variables
The `border-radius` variable list should be determined according to the baklava design system

[Figma](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=902%3A5202)

batuhantozun:
Border radius variables are defined to root css and changed static variables.
Closes #185

### DIFF
--- a/src/components/tab-group/bl-tab-group.css
+++ b/src/components/tab-group/bl-tab-group.css
@@ -7,9 +7,11 @@
 
 .tabs-list {
   overflow-x: scroll;
-  border-radius: var(--bl-border-radius-m) var(--bl-border-radius-m) var(--bl-border-radius-0) var(--bl-border-radius-0);
+  border-radius: var(--bl-border-radius-m) var(--bl-border-radius-m) var(--bl-border-radius-0)
+    var(--bl-border-radius-0);
 }
 
 .panels {
-  border-radius: var(--bl-border-radius-0) var(--bl-border-radius-0) var(--bl-border-radius-m) var(--bl-border-radius-m);
+  border-radius: var(--bl-border-radius-0) var(--bl-border-radius-0) var(--bl-border-radius-m)
+    var(--bl-border-radius-m);
 }
--- a/src/components/tab-group/bl-tab-group.css
+++ b/src/components/tab-group/bl-tab-group.css
@@ -7,12 +7,9 @@
 
 .tabs-list {
   overflow-x: scroll;
-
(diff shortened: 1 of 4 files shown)


src/themes/default.css
```diff
@@ -75,6 +75,15 @@
   --bl-font-size-5xl: 3rem; /* 48px */
   --bl-font-size-6xl: 4rem; /* 64px */
 
+  /* Border radius */
+  --bl-border-radius-0: 0;
```
muratcorlu, line 79: I'm not sure if we should have a variable for `0` value since this means no border-radius. We don't have 0 values in size list. I think it's better to not include this too.

muratcorlu (commented):
Can we also add a page to Storybook documentation under Design System?

leventozen (approved):
🚀

src/components/tab-group/bl-tab-group.css
```diff
@@ -7,12 +7,11 @@
...
-  /* FIXME: use variables */
-  border-radius: 6px 6px 0 0;
+  border-radius: var(--bl-border-radius-m) var(--bl-border-radius-m) var(--bl-border-radius-0)
+    var(--bl-border-radius-0);
 }
 
 .panels {
-  /* FIXME: use variables */
-  border-radius: 0 0 6px 6px;
+  border-radius: var(--bl-border-radius-0) var(--bl-border-radius-0) var(--bl-border-radius-m)
```
muratcorlu, line 15: Need to remove `--bl-border-radius-0` variable. 
```suggestion
  border-radius: 0 0 var(--bl-border-radius-m)
```

src/components/tab-group/tab-panel/bl-tab-panel.css
```diff
@@ -1,8 +1,7 @@
 :host {
   display: none;
-
-  /* FIXME: use variables */
-  border-radius: 0 0 6px 6px;
+  border-radius: var(--bl-border-radius-0) var(--bl-border-radius-0) var(--bl-border-radius-m)
```
muratcorlu, line 3:
```suggestion
  border-radius: 0 0 var(--bl-border-radius-m)
```

muratcorlu (changes requested):

muratcorlu:
@batuhantozun there are more places that we set custom border-radius values. Can you replace all of them?

PR #217, opened 2022-09-06
feat: alert component

koulgar:
w/ @canerturul 
This PR contains new alert component feature with following;

- Storybook file,
- Unit tests for component itself and converter util,
- New npm package (@storybook/addon-viewport) for storybook to display responsive design,


src/components/alert/bl-alert.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  shouldRender(value: undefined | boolean | string, html: TemplateResult) {
+    if (!value) return null;
+    return html;
+  }
+
+  predefinedIcons() {
+    switch (this.variant) {
+      case ('success'):
+        return 'check_fill';
```
leventozen, line 67: Expected indentation of 4 spaces but found 6 from verify action 🐰

src/components/alert/bl-alert.ts
```diff
@@ -0,0 +1,135 @@
...
+      case ('error'):
```
leventozen, line 68: Expected indentation of 4 spaces but found 6 from verify action 🐰

src/components/alert/bl-alert.ts
```diff
@@ -0,0 +1,135 @@
...
+        return 'close_fill';
```
leventozen, line 69: Expected indentation of 4 spaces but found 6 from verify action 🐰

src/components/alert/bl-alert.ts
```diff
@@ -0,0 +1,135 @@
...
+      default:
```
leventozen, line 70: Expected indentation of 4 spaces but found 6 from verify action 🐰

src/components/alert/bl-alert.ts
```diff
@@ -0,0 +1,135 @@
...
+        return this.variant;
```
leventozen, line 71: Expected indentation of 4 spaces but found 6 from verify action 🐰

src/components/alert/bl-alert.ts
```diff
@@ -0,0 +1,139 @@
...
+import { ButtonVariant, ButtonKind, ButtonSize } from '../button/bl-button';
+import { classMap } from 'lit/directives/class-map.js';
+
+export type AlertVariant = 'info' | 'warning' | 'success' | 'error';
+
+/**
+ * @tag bl-alert
+ * @summary Baklava Alert component
+ *
+ * @cssproperty --bl-alert-display - Sets the display property of button. Default value is 'inline-block'.
```
muratcorlu, line 17: I'm not sure if we need a variable like this. If this only sets the `display` of `:host` then it can be already set outside (host element is on the outside of the shadowdom and styles of the host element can be overriden by component user).

muratcorlu (changes requested):
Please set PR title as `feat: alert component` PR title will be the default commit message for this PR.

src/components/alert/icon.converter.ts
```diff
@@ -0,0 +1,11 @@
+import type { ComplexAttributeConverter } from 'lit';
+
+export const iconConverter = (): ComplexAttributeConverter<string | boolean> => {
+  return {
+    fromAttribute: (value: string): string | boolean => {
+      if (!value || value === 'true') return true;
```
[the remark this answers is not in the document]
koulgar: Issue here that we use `icon` attribute as `boolean` and also `string` to change default icon. Lit's default converter does not work with multiple converters, so we needed to use this and this converter requires single type and if we use `boolean` as value type we can not catch when usage is like `icon="download"`. For this case when we use attribute as `icon` this value does not work like when attribute type is `boolean`. 
These cases requires to use converter hence requires a type check when using `getIcon()` function.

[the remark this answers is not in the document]
muratcorlu: Ok, I got your point. I also just checked [Lit document about custom converters](https://lit.dev/docs/components/properties/#conversion-converter). So you mean, if we don't use this converter and user sets `<bl-alert icon>` then value of the `icon` attribute becomes a string `true`, right? Did you check `value` here can ever be string `true` or `false`?

If this is a generic way of having an attribute with `boolean | string` type, maybe we can move this to our utlities folder with a more generic name to make it reusable in other components.

src/components/alert/bl-alert.ts
koulgar: I've realized that i kinda misunderstood the concept of `:host`.
In current state of this component when `closable` button clicked component adds class `hidden` to main wrapper (which already has `alert` class) and sets `display: none;` to main wrapper, not the `host`.
It's better to set display of `host` to `none` as you've said but i couldn't manage to change hosts `display` value without using a reflected attribute (`:host[(hidden)]`). 
I've override `--bl-display` of `host` when component has `class="alert hidden"` but it seems like not working either.
I'm feeling like i'm missing the obvious 😄, would you mind helping me with an example for close button to change `host`'s visibility?

src/components/alert/bl-alert.ts
```diff
@@ -0,0 +1,139 @@
...
+    </span>`;
+
+    const classes = classMap({
+      alert: true,
+      hidden: this._hidden,
+    });
+
+    return html`
+      <div class=${classes}>
+        <div class="wrapper">
```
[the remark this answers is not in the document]
koulgar: This will be used when component is on responsive resolution to change position of action slot by setting `flex-direction` to `column`.

src/components/alert/bl-alert.ts
muratcorlu: I think it make sense to have a reflected attribute named `hidden` (or maybe `closed` is better to be aligned with `closable` attribute) and style it as 

```css
:host([closed]) {
   display: none;
}
```

With this way we can animate showing/hiding alert in the feature.

You can also add public `open()` and `close()` methods just sets the `closed` property accordingly. Our internal close button also can just call same `close()` method.

Does this make sense?

muratcorlu: Public methods can be useful for the consumer to programmatically show/hide an alert. Of course `closed` will be `false` by default, so `open()` only help once `closed` is `true`.

src/components/alert/icon.converter.ts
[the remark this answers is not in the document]
koulgar: > Ok, I got your point. I also just checked Lit document about custom converters. So you mean, if we don't use this converter and user sets <bl-alert icon> then value of the icon attribute becomes a string true, right? Did you check value here can ever be string true or false?

for this ( `<bl-alert icon>` ) kind of usage It actually becomes empty string, instead of string `true`, if attribute type (or converter value type) is set to `boolean` then it converts to boolean `true`. 
In converter that's why i've added `!value` because if user adds `icon` attribute on component it should work as `boolean` `true` and `!value` covers this section.

> If this is a generic way of having an attribute with boolean | string type, maybe we can move this to our utlities folder with a more generic name to make it reusable in other components.

I agree to that, i think it's highly possible that we'll face another scenario like this. But i think we might need `date` to `string` converters like Lit tutorials mension. So instead of `utilities` folder, we might consider to have another folder called `converters`. What do you think?

src/components/alert/bl-alert.ts
koulgar: Ok this makes sense, i'm refactoring component accordingly. Thank you!

src/components/alert/icon.converter.ts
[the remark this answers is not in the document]
koulgar: addition to that, if we've agreed on i'm signing this comment https://github.com/Trendyol/baklava/pull/217/files#r963506474 to resolved.

src/components/alert/bl-alert.ts
```diff
@@ -0,0 +1,139 @@
...
+      case 'success':
+        return 'check_fill';
+      case 'error':
+        return 'close_fill';
+      default:
+        return this.variant;
+    }
+  }
+
+  getIcon(): string | undefined {
```
koulgar, line 75: should this also be private ? @muratcorlu

src/components/alert/bl-alert.ts
koulgar: should public methods mentioned here be added to issue page? @muratcorlu

src/components/alert/bl-alert.ts
```diff
@@ -0,0 +1,139 @@
...
+      return this.predefinedIcons();
+    }
+    return this.icon;
+  }
+
+  initAlertActionSlot() {
+    const actionSlot = this._getAlertActionSlot;
+    if (!actionSlot) return;
+    if (actionSlot?.tagName !== 'BL-BUTTON')
+      throw new Error('Action slot must contain bl-button component as child!');
```
koulgar, line 87: @muratcorlu @leventozen 
While im searching an issue right after refactoring `initAlertActionSlot()` to listen `slotchange` event. 
Found exactly same issue that i've facing [here](https://github.com/open-wc/open-wc/issues/2124) and that thread mentions that throwing an error inside a web component is actually anti-pattern and displaying a message inside component could be a solution. 
What would you suggest for this case?

buseselvi (commented):
It seems the typography of the texts is not the rubik font that we used, can we fix it?

src/components/alert/icon.converter.ts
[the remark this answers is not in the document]
muratcorlu: As far as I understand, value in this function never can be string `true` or string `false`. If so, we don't need to run a logic over this potential.

We can re-organize folder structure in the future if we'll have more and more converters. For now, I think just utilities folder is good enough for making this a common helper that multiple components can use.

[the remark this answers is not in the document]
muratcorlu: > addition to that, if we've agreed on i'm signing this comment https://github.com/Trendyol/baklava/pull/217/files#r963506474 to resolved.

I don't think this is directly related with here. Let me write some more explanation there.

src/components/alert/bl-alert.ts
muratcorlu: Yes, public methods also should be documented in the issue and the storybook.

src/components/alert/bl-alert.ts
muratcorlu: OK, that looks ok for now. If removing action slot creates a big annoyance with some valid arguments, we can consider to allow other type of elements in the future.

src/components/alert/bl-alert.ts
```diff
@@ -0,0 +1,139 @@
...
+      case 'error':
+        return 'close_fill';
+      default:
+        return this.variant;
+    }
+  }
+
+  getIcon(): string | undefined {
+    if (!this.icon) return;
+    if (typeof this.icon === 'boolean') {
```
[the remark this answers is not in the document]
muratcorlu: `this.icon === true` already makes a type check (triple equals) and we want to show predefined icons only if the `icon` is `true`. So instead of checking the type, just checking the value with triple equals is more efficient.

src/components/alert/bl-alert.css
```diff
@@ -0,0 +1,103 @@
...
+  .close {
+    margin-top: var(--bl-size-2xs);
+  }
+  
+  .wrapper {
+    flex-direction: column;
+    align-items: flex-start;
+    margin-bottom: var(--bl-padding);
+  }
+}
```
muratcorlu, line 84: Please update this part according to the latest decision.

src/components/alert/bl-alert.css
```diff
@@ -0,0 +1,103 @@
+:host {
+  --bl-display: block;
```
muratcorlu, line 2: We don't need this variable at all. Use can just set this like:

```css
bl-alert {
    display: inline-block;
}
```

src/components/alert/bl-alert.css
```diff
@@ -0,0 +1,103 @@
...
+  justify-content: space-between;
+  background-color: var(--bl-main-bg-color);
+  color: var(--bl-color-content-primary);
+  box-shadow: 0 0 0 1px var(--bl-main-color);
+  border-radius: var(--bl-border-radius-s);
+  padding: 0 var(--bl-padding);
+}
+
+.description {
+  font-size: var(--bl-description-font-size);
```
muratcorlu, line 24: I think description has it's own `font` style. You should not need to set size and line-height manually.

src/components/alert/bl-alert.css
```diff
@@ -0,0 +1,103 @@
...
+  margin-bottom: var(--bl-size-4xs);
+}
+
+.actions {
+  display: flex;
+  align-items: center;
+}
+
+.caption + .description {
+  --bl-description-font-size: var(--bl-font-size-s);
```
muratcorlu, line 63: Use `--bl-font-caption-text` for the description font when a caption is provided. Otherwise use `--bl-font-title-3-regular`

src/components/alert/bl-alert.css
```diff
@@ -0,0 +1,103 @@
...
+  --bl-display: block;
+
+  display: var(--bl-display);
+}
+
+.alert {
+  --bl-padding: var(--bl-size-m);
+  --bl-main-color: var(--bl-color-primary);
+  --bl-main-bg-color: var(--bl-color-accent-primary-background);
+  --bl-description-font-size: var(--bl-font-size-m);
```
muratcorlu, line 11: You don't even need to use `bl-` prefixes for those variables.

src/components/alert/bl-alert.ts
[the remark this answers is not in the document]
koulgar: oh, now i got it. Since the converter handles it, we don't have to check it here. I'm assuming that after i've added the converter, i've just forgot to refactore here. Thanks!

src/components/alert/bl-alert.stories.mdx
```diff
@@ -0,0 +1,163 @@
...
+</Canvas>
+
+### Captionless Alert
+
+Alerts don't always have to have a caption, you can use plain text to explain the situation.
+
+<Canvas isColumn>
+  <Story name="Captionless Info Alert">
+    {SingleAlertTemplate({content: "Informational Text", icon: true})}
+  </Story>
```
muratcorlu, line 109: With this approach, Storybook controls will not work. Do this instead:

```suggestion
  <Story name="Captionless Info Alert" args={{content: "Informational Text", icon: true}}>
    {SingleAlertTemplate.bind({})}
  </Story>
```

muratcorlu (approved):
Good job @koulgar, @canerturul !

I think we are good to go!

leventozen:
Great work all of you! 🚀 

I will merge this pr with eslint configurations. We may need an extra development for that part but not in the scope of this pr.

PR #225, opened 2022-09-08
feat: select component

mustafafrw:
This PR is for a new component: Select

Contains following features

- bl-select component
- bl-select-option component
- Storybook files
- Unit tests

Update: bl-checkbox component added into bl-select-option


src/components/select/bl-select.css
```diff
@@ -0,0 +1,221 @@
+:host {
+  --bl-select-padding-vertical: var(--bl-size-2xs);
```
leventozen, line 2: To apply safer approach that will prevent developers to style many components unintentionally, we should create a class (ex. `bl-select`) and define variables under this class. To hear more about this, you can look at [this](https://github.com/Trendyol/baklava/discussions/164) discussion.

src/components/select/bl-select.ts
```diff
@@ -0,0 +1,394 @@
...
+  }
+
+  private get _showPlaceHolder() {
+    if (this.label && !this.labelFixed) {
+      return !this._selectedOptions.length && this._isMenuOpen;
+    }
+    return !this._selectedOptions.length;
+  }
+
+  private get _isRemoveIconShow() {
```
muratcorlu, line 256: I suggest to use a different name. Like `isRemoveIconVisible` or `showRemoveIcon`

src/components/select/option/bl-select-option.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    this.updateComplete.then(() => {
+      const el = this.closest('bl-select');
+      // FIXME: We should warn when parent is not bl-select
+
+      el?.registerOption(this);
```
muratcorlu, line 67: Please also unregister this option in a `disconnectedCallback`. Options can dynamically be added/removed inside same `bl-select` instance.

src/components/select/option/bl-select-option.ts
```diff
@@ -0,0 +1,76 @@
...
+        <slot></slot>
+      </span>
+    </div> `;
+  }
+
+  onClickOption() {
+    this._onSelect({
+      value: this.value,
+      text: this.textContent,
+    } as ISelectOption);
```
muratcorlu, line 57: I think we also need to have `selected` information here. Currently, in multi selection mode, "selecting" and "unselecting" events will send same data.

src/components/select/option/bl-select-option.ts
```diff
@@ -0,0 +1,76 @@
...
+  disabled?: boolean = false;
+
+  /**
+   * Sets option as selected state
+   */
+  @property({ type: Boolean, reflect: true })
+  selected?: boolean = false;
+
+  @state()
+  isCheckbox?: boolean = false;
```
muratcorlu, line 33: There is a problem here but I need some more time to explain it 😊

I tried, I wrote many things but then removed all. First I need to evaluate some options on my mind. I'll come back here soon.

muratcorlu (commented):
Good job @mustafafrw!

Some parts are more complicated than I expected. I shared my partial review notes. I will continue soon about other parts.

muratcorlu:
We are suffering with linter errors in every PRs. There is something wrong on our linter configuration for sure. We really need to look at it deeply @leventozen :(

leventozen:
> We are suffering with linter errors in every PRs. There is something wrong on our linter configuration for sure. We really need to look at it deeply @leventozen :(
😕 today, I will work on it. I think I found out what is the issue.

src/components/select/bl-select.css
```diff
@@ -0,0 +1,218 @@
...
+  --bl-select-icon-color: var(--bl-color-content-tertiary);
+  --bl-select-text-color: var(--bl-color-content-primary);
+  --bl-select-label-color: var(--bl-color-content-secondary);
+  --bl-select-placeholder-color: var(--bl-color-content-tertiary);
+  --bl-select-height: var(--bl-size-2xl);
+  --bl-select-menu-padding: 0 var(--bl-size-m);
+  --bl-select-menu-margin-top: var(--bl-size-2xs);
+  --bl-select-font-size: var(--bl-font-size-m);
+  --bl-select-disabled-color: var(--bl-color-tertiary);
+  --bl-select-menu-height: 250px;
```
muratcorlu, line 26: No need to use `bl-select` prefixes in local variables. Since those variables can only be used inside this component, just use variables without prefixes. Like `--padding-vertical`

src/components/select/option/bl-select-option.css
```diff
@@ -0,0 +1,61 @@
+:host {
+  --bl-select-option-font: var(--bl-font-title-3-regular);
+  --bl-select-option-padding: var(--bl-size-xs) 0;
+  --bl-select-option-selected-color: var(--bl-color-primary);
+  --bl-select-option-hover-color: var(--bl-color-primary-hover);
+  --bl-select-option-color: var(--bl-color-secondary);
+  --bl-select-option-disabled-color: var(--bl-color-secondary);
+  --bl-select-option-seperator: 1px solid var(--bl-color-border);
+  --bl-select-option-gap: var(--bl-size-2xs);
+  --bl-select-option-transition: color 120ms ease-out;
```
muratcorlu, line 10: Same rules we mentioned for the `bl-select` also applies here.

src/components/select/bl-select.css
```diff
@@ -0,0 +1,218 @@
+:host {
+  width: 200px;
+  display: inline-block;
+
+  --bl-select-menu-z-index: 1;
```
muratcorlu, line 5: Do we need to give this property outside? If so, it should be documented in the component file. But I think we should have a standard about z-levels and use them in the components accordingly. The components like tooltip, select, date-picker, dialog, drawer will need a consistent vertical levelling. @buseselvi

src/components/select/bl-select.ts
```diff
@@ -0,0 +1,402 @@
...
+        <bl-icon
+          name="${this._isMenuOpen ? 'arrow_up' : 'arrow_down'}"
+          title="arrow_down"
+          style="font-size: var(--bl-font-size-m)"
+        ></bl-icon>
+      </div>
+    </div>`;
+  }
+
+  menuTemplate() {
```
muratcorlu, line 215: This should be private.

src/components/select/bl-select.ts
```diff
@@ -0,0 +1,402 @@
...
+
+    document.addEventListener('click', this._clickOutsideHandler);
+  }
+  disconnectedCallback() {
+    super.disconnectedCallback();
+
+    document.removeEventListener('click', this._clickOutsideHandler);
+  }
+
+  inputTemplate() {
```
muratcorlu, line 177: This should be private.

src/components/select/bl-select.ts
```diff
@@ -0,0 +1,402 @@
...
+      strategy: 'absolute',
+      middleware: [
+        flip(),
+        offset(8),
+        size({
+          apply(args: MiddlewareArguments & { availableHeight: number }) {
+            Object.assign(args.elements.floating.style, {
+              maxHeight: `${
+                args.availableHeight > defaultMaxHeight ? defaultMaxHeight : args.availableHeight
+              }px`,
```
muratcorlu, line 155: Why we are not just setting a `max-height` in CSS?

src/components/select/bl-select.ts
```diff
@@ -0,0 +1,402 @@
...
+      ${this._selectedOptions.map(item => html`<li>${item.text}</li>`)}
+    </ul>`;
+    const _selectedItemCount = this._additionalSelectedOptionCount
+      ? html`<span>+${this._additionalSelectedOptionCount}</span>`
+      : null;
+    const removeIcon = this._isRemoveIconVisible
+      ? html`<bl-icon
+          class="remove-all"
+          name="close"
+          title="close"
```
muratcorlu, line 188: `title` is not needed here.

src/components/select/bl-select.ts
```diff
@@ -0,0 +1,402 @@
...
+          style="font-size: var(--bl-font-size-xs);"
```
muratcorlu, line 189: Avoid using inline styles.

src/components/select/bl-select.ts
```diff
@@ -0,0 +1,402 @@
...
+
+  inputTemplate() {
+    const inputSelectedOptions = html`<ul class="selected-options">
+      ${this._selectedOptions.map(item => html`<li>${item.text}</li>`)}
+    </ul>`;
+    const _selectedItemCount = this._additionalSelectedOptionCount
+      ? html`<span>+${this._additionalSelectedOptionCount}</span>`
+      : null;
+    const removeIcon = this._isRemoveIconVisible
+      ? html`<bl-icon
```
muratcorlu, line 185: We should use `bl-button` for icon only buttons. secondary text button with an `icon` and `label` attribute.

src/components/select/bl-select.ts
```diff
@@ -0,0 +1,402 @@
...
+      ?disabled=${this.disabled}
+      @click=${this._onClickSelectInput}
+    >
+      ${placeholder} ${inputSelectedOptions} ${_selectedItemCount}
+      <div class="actions">
+        ${removeIcon}
+        <bl-icon
+          name="${this._isMenuOpen ? 'arrow_up' : 'arrow_down'}"
+          title="arrow_down"
+          style="font-size: var(--bl-font-size-m)"
```
muratcorlu, line 209: Avoid using inline styles.

src/components/select/bl-select.ts
```diff
@@ -0,0 +1,402 @@
...
+      class="select-input"
+      ?disabled=${this.disabled}
+      @click=${this._onClickSelectInput}
+    >
+      ${placeholder} ${inputSelectedOptions} ${_selectedItemCount}
+      <div class="actions">
+        ${removeIcon}
+        <bl-icon
+          name="${this._isMenuOpen ? 'arrow_up' : 'arrow_down'}"
+          title="arrow_down"
```
muratcorlu, line 208: `title` is not needed here.

buseselvi:
Can we fix the Disabled SelectOption color to Content Passive (#AFBBCA) color? @mustafafrw

src/components/select/bl-select.css
mustafafrw: I think there should be a way to customize a component intentionally. I moved these styles under a class but in my opinion z-index property should stay under :host because developers might need to change it.

mustafafrw:
> Can we fix the Disabled SelectOption color to Content Passive (#AFBBCA) color? @mustafafrw



> Can we fix the Disabled SelectOption color to Content Passive (#AFBBCA) color? @mustafafrw

Done 🚀

src/components/select/option/bl-select-option.ts
mustafafrw: I refactored the code. In this mode, bl-select-option handles the `selected` status and emit to parent, I added selected status into event detail.

src/components/select/bl-select.ts
```diff
@@ -0,0 +1,394 @@
...
+   * Shows the component in error state
+   */
+  @property({ type: Boolean })
+  error?: boolean = false;
+
+  /**
+   * Shows the component in success state
+   */
+  @property({ type: Boolean })
+  success?: boolean = false;
```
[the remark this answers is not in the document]
mustafafrw: You are perfectly right these are open questions and need some time to clarify. But when I saw it in Figma design I thought I should do it. For such cases, we should consider enter more details in issue (cases that in Figma design but we will not do for now can be written inside the issue)

src/components/select/bl-select.css
mustafafrw: I think we should give this property outside. This is why I didn't remove z-index property from :host. In some cases it might need to be edited.

src/components/select/bl-select.ts
mustafafrw: I tried with these attributes but I couldn't get the exact appearance 😿

src/components/select/bl-select.ts
```diff
@@ -0,0 +1,394 @@
...
+      this._isInvalid = this._selectedOptions.length === 0;
+    }
+  }
+
+  protected updated(_changedProperties: PropertyValues) {
+    if (
+      _changedProperties.has('_selectedOptions') &&
+      _changedProperties.get('_selectedOptions') instanceof Array
+    ) {
+      this._checkAdditionalItemCount();
```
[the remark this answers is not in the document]
mustafafrw: We know items checked status but we don't how many selected items visible in input area. We should wait till the end of the update cycle and count the visible elements. This way we can give the number of elements outside the visible area. 

We only need to call this method when bl-select is `multiple`. I added a condition here.

src/components/select/bl-select.ts
mustafafrw: Let's think a container there is not any available space below and has limited space above. If we don't limit the max height to available height, some options will not visible as seen in the example below. 

set `max-height` to available height
image
only `max-height` in css
image

buseselvi:
Can we also fix the Disabled Selected Option color to Content Passive (#AFBBCA) color? 🙏🏻 @mustafafrw 
image

mustafafrw:
> Can we also fix the Disabled Selected Option color to Content Passive (#AFBBCA) color? 🙏🏻 @mustafafrw image

Done 🚀

src/components/select/bl-select.css
```diff
@@ -0,0 +1,224 @@
...
+  justify-content: center;
+  gap: var(--bl-size-2xs);
+  margin-left: var(--bl-size-2xs);
+}
+
+.select-menu {
+  --left: 0;
+  --top: 0;
+
+  position: absolute;
```
muratcorlu, line 138: When we develop tooltip component, we had issues about absolute positioning if container of the element has a limited height with `overflow: hidden` then we decided to use ‘fixed’ positioning. Maybe we can consider same approach for the consistency.

src/components/select/bl-select.css
muratcorlu: I would consider to have static value for now and convert it to the level that we’ll define with our design token later. In that case, let’s say we have a `—bl-level-high: 10;` variable and use it here, and user wants to change it, then they can override `—bl-level-high` in whole page or in a scope, but we can still guarantiee that our other components those uses same level will continue to work consistently with select component. But if a user sets the `z-index` of a specific select, then there will be a risk to have conflict with our other popover components like tooltip, date-picker, dialog, drawer etc).

src/components/select/bl-select.ts
muratcorlu: If I understand correctly, you mean this is related using absolute position relative to parent. Maybe we need to consider using fixed positioning (as I mentioned in another thread). In that case you may re-consider here to simplify logic.

src/components/select/bl-select.ts
muratcorlu: Let me also have a look for that.

src/components/select/bl-select.css
mustafafrw: Yes, we have issues with absolute positioning, they will be resolved when we change the positioning to `fixed`. But when we move to fixed positioning, we will have other problems. For example, during scrolling, the dropdown menu will appear in places we don't want. How do we find a solution to this situation? I think it will be more difficult to apply the solution in tooltip to select component.

src/components/select/bl-select.css
mustafafrw: Alright, that solution will be better and more efficient. I will remove this value from `:host`

src/components/select/bl-select.ts
mustafafrw: Note: We found a solution for fixed positioning problems. We changed positioning to `fixed` with the help of autoUpdate function in floating-ui package.

muratcorlu (approved):
Good job @mustafafrw!

I think currently only big missing part is using select as a form field. We'll need ElementInternals integration with `name` and `value` attributes, but I think it's ok to merge this for now and track that part in a separate issue. @leventozen, does it make sense to you, too?

buseselvi (approved):

PR #237, opened 2022-09-14
docs(design): add sizing to design system

olkeoguz:
This pl includes a sizing doc:

- To show the sizing variables in baklava with their values and,
- To illustrate the usage

### DIFF
--- a/docs/design-system/sizing.stories.mdx
+++ b/docs/design-system/sizing.stories.mdx
@@ -6,19 +6,21 @@ import { Meta } from '@storybook/addon-docs';
 
 Baklava uses a list of defined size variables.Using the right sizes will create consistency and hierarchy in user interface.
 
+In our provided themes, we are using `rem` values for the sizing variables to provide better accessibility. You can see pixel equivalents in the table below, with default browser sizing (1rem = 16px)
+
 | Variable                |      Value      |   <div style={{ width: '25rem' }}>Example</div>  |
 |-----------------------|:-------------:|:----------:|
-| --bl-size-4xs |  0.125rem (2px)   | <div style={{ width: '5rem', height: '0.125rem', backgroundColor: '#d9e0e8' }}></div> |
-| --bl-size-3xs |  0.25rem  (4px)   | <div style={{ width: '5rem', height: '0.25rem', backgroundColor: '#d9e0e8' }}></div> |
-| --bl-size-2xs |  0.5rem  (8px)    | <div style={{ width: '5rem', height: '0.5rem', backgroundColor: '#d9e0e8' }}></div> |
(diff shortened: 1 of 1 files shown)


docs/design-system/sizing.stories.mdx
```diff
@@ -0,0 +1,24 @@
...
+
+<Meta title="Design System/Size and Spacing" />
+
+# Size and Spacing
+
+Baklava uses a list of defined size variables.Using the right sizes will create consistency and hierarchy in user interface.
+
+| Variable                |      Value      |   <div style={{ width: '25rem' }}>Example</div>  |
+|-----------------------|:-------------:|:----------:|
+| --bl-size-4xs |  0.125rem (2px)   | <div style={{ width: '5rem', height: '0.125rem', backgroundColor: '#d9e0e8' }}></div> |
```
muratcorlu, line 11: Can we use variables here for setting Example height? Like `height: 'var(--bl-size-4xs)'`

docs/design-system/sizing.stories.mdx
```diff
@@ -0,0 +1,24 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Design System/Size and Spacing" />
+
+# Size and Spacing
+
+Baklava uses a list of defined size variables.Using the right sizes will create consistency and hierarchy in user interface.
+
```
muratcorlu, line 8: I think it'll be nice to do an addition like below:

> In our provided themes, we are using `rem` values for the sizing variables to provide better accessibility. You can see pixel equivalents in the table below, with default browser sizing (1rem = 16px)

docs/design-system/sizing.stories.mdx
```diff
@@ -0,0 +1,24 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Design System/Size and Spacing" />
+
+# Size and Spacing
+
+Baklava uses a list of defined size variables.Using the right sizes will create consistency and hierarchy in user interface.
```
muratcorlu, line 7:
```suggestion
Baklava uses a list of defined size variables. Using the right sizes will create consistency and hierarchy in the user interface.
```

docs/design-system/sizing.stories.mdx
muratcorlu: It would be great to generate this table programmatically. So we could be sure it'll always udpated and no repeating works... Maybe in the future 😊

leventozen (approved):

docs/design-system/sizing.stories.mdx
olkeoguz: We'll be able to do it once we release the bl-data-grid maybe ? 🤞

PR #238, opened 2022-09-14
docs: add how to customize a components style

leventozen:
Document of the rules that we discussed at [here](https://github.com/Trendyol/baklava/discussions/164).


muratcorlu:
I would expect to touch on those points in this document:

* Intentional limited flexibility regarding the nature of the Design System (instead of a UI Library)
* Themeable Design Tokens to customize all of the components for project's own branding, design language
* Using CSS Custom Properties for customizing specific component styles. ShadowDOM and the nature of web components about styling
* For Contributors:
  * Which parts of the components should be styleable outside
  * How to give styling options to consumers of the component. Prefixing custom properties with component's name
  * Styling of `:host` element and using private variables on `:host`

leventozen:
> I would expect to touch on those points in this document:
> 
> * Intentional limited flexibility regarding the nature of the Design System (instead of a UI Library)
> * Themeable Design Tokens to customize all of the components for project's own branding, design language
> * Using CSS Custom Properties for customizing specific component styles. ShadowDOM and the nature of web components about styling
> * For Contributors:
>   
>   * Which parts of the components should be styleable outside
>   * How to give styling options to consumers of the component. Prefixing custom properties with component's name
>   * Styling of `:host` element and using private variables on `:host`

Thanks for this suggestion. Seems like it would be better with that information. Love the structure as well. I will write those missing parts.

olkeoguz (approved):

PR #241, opened 2022-09-20
docs: playground usage

### LINKED ISSUE #228: Documentation - Playground usage

leventozen:
In this pr, I will try to explain why and how we use `playground`.

Closes #228


docs/how-to-use-playground.stories.mdx
```diff
@@ -0,0 +1,26 @@
...
+Playground is a demonstration area for developers. It stays under the `playground` folder in the root. It helps you to test your features & components by eliminating the need for stories.
+
+<bl-alert variant="warning" caption="Note">You should not commit the changes in the playground folder.</bl-alert>
+
+You can run it with `npm run serve`, then it builds project and run the build in `watch mode` at port `8000`. You can either create a `foo.html` file and work on it or simply use `template.html`.
+It looks like this,
+
+<img src={playgroundLandingPage}></img>
+
+Every change in your `html` file will be served immediately. However, if you change your component, you should rebuild the project by using `npm run build`.
```
muratcorlu, line 26: > However, if you change your component, you should rebuild the project by using `npm run build`.

That's not needed in this mode. `npm run serve` already works on watch mode for whole build process.

docs/how-to-use-playground.stories.mdx
```diff
@@ -0,0 +1,26 @@
...
+    },
+  }}
+/>
+
+# How to use playground?
+Playground is a demonstration area for developers. It stays under the `playground` folder in the root. It helps you to test your features & components by eliminating the need for stories.
+
+<bl-alert variant="warning" caption="Note">You should not commit the changes in the playground folder.</bl-alert>
+
+You can run it with `npm run serve`, then it builds project and run the build in `watch mode` at port `8000`. You can either create a `foo.html` file and work on it or simply use `template.html`.
```
muratcorlu, line 21: I think we should suggest starting with copying `template.html` file as `foo.html` and work on it. As `template` name comes from this intention. In this case user will not need to extra attention to not commit changes accidentally, since `playground` folder is already ignored from git, except `template.html` file.

docs/how-to-use-playground.stories.mdx
```diff
@@ -0,0 +1,26 @@
...
+        hidden: true,
+      },
+    },
+  }}
+/>
+
+# How to use playground?
+Playground is a demonstration area for developers. It stays under the `playground` folder in the root. It helps you to test your features & components by eliminating the need for stories.
+
+<bl-alert variant="warning" caption="Note">You should not commit the changes in the playground folder.</bl-alert>
```
muratcorlu, line 19:
```suggestion
<bl-alert variant="warning" caption="Note">playground folder is ignored from Git index except template.html file. It's not intended to make changes on template.html file unless you want to suggest something on template that you consider it will be helpful for everyone.</bl-alert>
```

And maybe we can move this alert below the next paragraph.

docs/how-to-use-playground.stories.mdx
leventozen: 🤔

muratcorlu (approved):

PR #253, opened 2022-09-28
fix(tooltip): use max-content width for tooltip content

Enes5519:
The tooltip shrink without off max-content width. You can see from [Stackblitz](https://stackblitz.com/edit/vitejs-vite-pdvgcp?file=src/App.css).

Steps to reproduce:
1. Create a fixed size div with `transform: translateY(1px);` style.
2. Add tooltip inside.

OLD:
image

NEW:
image

### DIFF
--- a/src/components/tooltip/bl-tooltip.css
+++ b/src/components/tooltip/bl-tooltip.css
@@ -22,6 +22,7 @@
   visibility: var(--visibility);
   z-index: 999;
   max-width: 424px;
+  width: max-content;
 }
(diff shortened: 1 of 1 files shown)


DamlaDemir (approved):

muratcorlu:
Currently UI tests fails because of an issue on select component that I created a fix with #256. Once it's merged, I expect UI tests will be successful here too.

leventozen (approved):

PR #257, opened 2022-09-29
fix(select): popover and selected issues

### LINKED ISSUE #252: bl-select's popover cannot be opened
**Describe the bug**
When the bl-select component is used inside the render method of another baklava component, the popover is not opening where the options should be listed.

**To Reproduce**
1. Create a component.
2. Return the bl-select element with some options.
3. Click to open the dropdown.
4. The _isPopoverOpen state won't be set to true.

Demo:
https://codesandbox.io/s/litelement-forked-h5c55o?file=/src/index.js

**Expected behavior**
The popover should be opened and closed by clicking inside the component.

**Additional context**
The problem does not occur when the bl-select is used outside of shadowRoot of another component.(i.e: inside the html body).

For this reason,a quick fix can be done in the "_clickOutsideHandler" function of bl-select by changing it as:

```ts
 const target = event.target as HTMLElement;
    const containsTarget = target.tagName === 'BL-SELECT' ? target.contains(this) :
      target.shadowRoot?.querySelector('bl-select')?.contains(this);

    if (!containsTarget && this._isPopoverOpen) {
      this.close();
      this._checkRequired();
    }
```
 
@mustafafrw  @muratcorlu  what do you think?

mustafafrw:
- As mentioned in this [issue#252](https://github.com/Trendyol/baklava/issues/252) there is a bug in `_closeOutsideHandler` function. 
- I realized another bug when `multiple = false`. It is not appearing in the selected state when an option is selected.
- Also, event listeners moved into `open` and `close` functions

Fixes #252

### DIFF
--- a/src/components/select/bl-select.ts
+++ b/src/components/select/bl-select.ts
@@ -133,17 +133,19 @@ export default class BlSelect extends LitElement {
   open() {
     this._isPopoverOpen = true;
     this._setupPopover();
+    document.addEventListener('click', this._clickOutsideHandler);
   }
 
   close() {
     this._isPopoverOpen = false;
     this._cleanUpPopover && this._cleanUpPopover();
+    document.removeEventListener('click', this._clickOutsideHandler);
   }
 
   private _clickOutsideHandler = (event: MouseEvent) => {
-    const target = event.target as HTMLElement;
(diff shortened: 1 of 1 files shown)


muratcorlu (approved):
Good job! 👏🏻

olkeoguz (approved):

PR #270, opened 2022-10-03
feat: purge cache on cdn

### LINKED ISSUE #262: Purging jsDelivr caches on releases
In this issue, we will add a purging cache mechanism after every release on jsDelivr.

leventozen:
Added purging cache mechanism after release.

Closes #262.

### DIFF
--- a/.github/workflows/purge-cache.yml
+++ b/.github/workflows/purge-cache.yml
@@ -1,0 +1,16 @@
+name: Purge Cache
+
+on:
+  workflow_call:
+    secrets:
+      SONAR_TOKEN:
+        required: true
+
+jobs:
+  purge-jsdelivr-cache:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: gacts/purge-jsdelivr-cache@v1
+        with:
+          url: |
+            https://cdn.jsdelivr.net/npm/@trendyol/baklava@beta
--- a/.github/workflows/purge-cache.yml
+++ b/.github/workflows/purge-cache.yml
@@ -2,9 +2,6 @@ name: Purge Cache
 
 on:
   workflow_call:
-    secrets:
-      SONAR_TOKEN:
-        required: true
 
 jobs:
   purge-jsdelivr-cache:
--- a/.github/workflows/verify-and-release.yml
+++ b/.github/workflows/verify-and-release.yml
@@ -23,3 +23,7 @@ jobs:
     uses: trendyol/baklava/.github/workflows/release.yml@next
     needs: verify
     secrets: inherit
+  purge-cache:
+    uses: trendyol/baklava/.github/workflows/purge-cache.yml@next
+    needs: release
+    secrets: inherit


.github/workflows/purge-cache.yml
```diff
@@ -0,0 +1,13 @@
...
+  workflow_call:
+
+jobs:
+  purge-jsdelivr-cache:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: gacts/purge-jsdelivr-cache@v1
+        with:
+          url: |
+            https://cdn.jsdelivr.net/npm/@trendyol/baklava@beta
```
muratcorlu, line 13: Purging that URL will not be enough. I asked this to jsDelivr before:

> Murat: Thank you for your response. Can we purge a prefix/alias like `/npm/@trendyol/baklava@beta/`? Or do we need to purge cache for every files specifically?
> Dmitriy: You would have to purge every file specifically. Most of our CDNs don't support prefix purging unfortunately.

So we need to purge every file url on `@beta` channel that has same name like `package.json`, `baklava.js`, `baklava-react.js`, `customElements.json`, all of the component entrance files (like `dist/components/button/bl-button.js`) and SVG files. :(

I think it'll be wise to find the URLs of the changed files only (especially for the SVGs)

.github/workflows/purge-cache.yml
```diff
@@ -0,0 +1,13 @@
+name: Purge Cache
+
+on:
+  workflow_call:
+
+jobs:
+  purge-jsdelivr-cache:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: gacts/purge-jsdelivr-cache@v1
```
muratcorlu, line 10: I checked the source code of this action and noticed that it makes separate requests for every URL purge. But there is also a POST endpoint to purge multiple caches by a single request. Considering purging caches for many files, I think we should use batch endpoint.

> curl -X POST \
  [http://purge.jsdelivr.net/](https://clicktime.symantec.com/15t5ZrVW89yGfvgAzx8Mu?h=5qY1Lcx4f2-Dseq_AtTbDXbpznpzdQwFTsIPKUdk3TI=&u=http://purge.jsdelivr.net/) \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json' \
  -d '{
"path": [
"/npm/test",
"/npm/test2",
"/npm/test3"
]
}'

PS: It works with up to 20 paths.

These are the informations that I got from jsDelivr by e-mail.

muratcorlu (changes requested):
I think current version will not work. I commented on the issues.

.github/workflows/purge-cache.yml
leventozen: 🤔 thanks for the comment. then we need to create an action to use in our repository to be able to purge and have a more trustworthy solution.

.github/workflows/purge-cache.yml
leventozen: 😞

leventozen:
I will create a github action for purge jsDelivr CDN in another project. So, I close this right now.

Issue #250, opened 2022-09-27
Refactoring button variants


batuhantozun:
In this issue, we'll refactor our button variants as a discussed in https://github.com/Trendyol/baklava/discussions/216

Our variants will be;

- primary
- secondary
- tertiary

Our kinds will be;

- default
- neutral
- success
- danger

| kind \ variant  | `primary` | `secondary` | `tertiary` |
| -------------- | --------- | ------------- | --------- |
| `default` | <img width="146" alt="image" src="https://user-images.githubusercontent.com/127687/188454809-601849c5-5d9a-421e-a68c-afd331663ae3.png"> | <img width="164" alt="image" src="https://user-images.githubusercontent.com/127687/188455371-41375fab-d129-4ee6-937c-01247df3f0d4.png"> | <img width="144" alt="image" src="https://user-images.githubusercontent.com/127687/188455455-330be0bf-3b26-40ad-b9a8-7db2cd24dce4.png"> |
| `neutral` | <img width="149" alt="image" src="https://user-images.githubusercontent.com/127687/188455648-502ae73c-a27a-4174-a1ff-f2aca09a984e.png"> | <img width="169" alt="image" src="https://user-images.githubusercontent.com/127687/188455722-8e6419fb-80fe-4d2f-9373-b77c0fe8b714.png"> | <img width="145" alt="image" src="https://user-images.githubusercontent.com/127687/188455774-5bb904ee-d83c-4ef1-819e-b42f784d7396.png"> |
| `success` | <img width="149" alt="image" src="https://user-images.githubusercontent.com/127687/188455890-5352fc60-18be-4138-b5c2-ecae1df33531.png"> | <img width="166" alt="image" src="https://user-images.githubusercontent.com/127687/188455926-cfc7e11d-a9fb-458b-ba9c-c8a48bd26785.png"> | <img width="139" alt="image" src="https://user-images.githubusercontent.com/127687/188455964-6e5c8f47-d0b0-4dbb-9cf2-f5d7c0858995.png"> |
[...]

PR #271, opened 2022-10-04
feat: refactor button variants

olkeoguz:
Includes refactoring button variants discussed in https://github.com/Trendyol/baklava/issues/250

Closes [#250](https://github.com/Trendyol/baklava/issues/250).

### DIFF
--- a/src/components/button/bl-button.css
+++ b/src/components/button/bl-button.css
@@ -2,7 +2,9 @@
   display: var(--bl-button-display, inline-block);
   max-width: 100%;
   position: relative;
+}
 
+.button {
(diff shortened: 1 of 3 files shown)


muratcorlu (changes requested):
Some other components that uses button inside should also be updated (Alert, Select, Tab)

muratcorlu:
Latest changes causes an alignment issue on Select component. (border next to cancel button)

image

muratcorlu (approved):
Looks good! 🚀

muratcorlu:
Only thing is, this PR should include `BREAKING CHANGE` in merge commit message since it's changing current structure of the button component.

Issue #91, opened 2022-04-06
Design System - Colors


muratcorlu:
### Design:

[Figma Design Document](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=28%3A2679)

### We'll implement this features on this issue;

- Palette colors will be named semantically
- Colors are conform with accessibility standards
- Colors should calculate mathematically. For example [here](https://toolness.github.io/accessible-color-matrix/)

muratcorlu:
During our investigations, together with @buseselvi and @fatihhayri we concluded with a plan:

* Although HSL is known as not a perfect fit for creating a consistent, accessible color palette, because of the lack of browser support of some newer color functions in CSS (like LCH) we decided to use HSL anyway to have a more consistent (comparing to HEX/RGB colors) color coding in our color pallettes.
* We'll update our current color pallette with a list of new colors those passing APCA standards.
* We decided to use APCA over WCAG2.1 because current WCAG standard is already known as not consistent about color contrast in some different use cases. Altough APCA is not a standard yet, it's calcualtion takes the consideration of background/foreground colors and perceptual contrast calculation.
* We'll create a PR with new color definitions with HSL and share some light value of the main color variants like key, hover, background.
* We noticed that we already have some wrong usage of some similar colors (mostly secondary and content colors) while playing with new color palette. Those will be also fixed while implementing new pallette.
* With a better documentation about theming Baklava, we'll focus on creating own color palettes for the projects that uses Baklava. Provided color palette will be a default, accessible palette but people will be free to use current or any other colors if they want.

We made a playground on codepen to test contrast values and their end results on Baklava components: https://codepen.io/muratcorlu/pen/MWXNNMp

## Resources:
* MDN CSS Color documentation: https://developer.mozilla.org/en-US/docs/Web/CSS/color
* A Guide To Modern CSS Colors With RGB, HSL, HWB, LAB And LCH: https://www.smashingmagazine.com/2021/11/guide-modern-css-colors/
* Color.js contrast calculator documentation: https://colorjs.io/docs/contrast.html
* An online contrast calculator: https://huetone.ardov.me
* Lyft Color System: https://design.lyft.com/re-approaching-color-9e604ba22c88
* Perceptually uniform color spaces: https://programmingdesignsystems.com/color/perceptually-uniform-color-spaces/index.html
* A presentation about OKLCH: https://evilmartians.com/events/oklch
[...]

PR #277, opened 2022-10-12
fix: featured and alternative colors fixed and new colors added

DamlaDemir:
Featured and alternative hover colors were wrong, they were fixed and added new content color because we were using the `--bl-color-primary-background` for white content text but it doesn't fit our theme rules as semantically.


buseselvi (approved):

olkeoguz (approved):

muratcorlu:
I have many questions but don't want to block this change now. I think we need to prioritize #91 We don't have a clear document and plan yet about our color palette. Maybe we can start working on it next week? @buseselvi

muratcorlu (approved):

Issue #141, opened 2022-06-27
Notification Component


muratcorlu:
## Design
[Figma Design Document ](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=80%3A3131&t=mOa06L4RJiSCbIvJ-0)

## `bl-notification` Implementation
General usage example:

```html
<bl-notification
    no-animation
    duration="7"
></bl-notification>

<script>
  const el = document.querySelector("bl-notification");

  const addedNotification = el.addNotification({
    caption: "Notification Caption",
    description: "This is a notification",
    variant: "warning",
    icon: "academy",
    primaryAction: {
      label: "Action",
      onClick: notification => {
        notification.remove();
      },
    },
    secondaryAction: {
      label: "Action",
      onClick: async notification => {
        await notification.remove();
        first.remove();
      },
    },
  });

  // some actions
  addedNotification.remove();
  // or
  el.removeNotification(addedNotification.id);
</script>
```

### Rules

* `duration` attribute sets the default duration of notifications when not provided.
* `no-animation` attribute disables animation of notifications. Animations will respect the user's preferences regardless of this property.
* Cards animate in from right on desktop and animate in from top on mobile (screens smaller than 480px).
* Last in card will be on top of the notification list on desktop, and last in card will be on bottom of the notification list on mobile.
* Cards has touch support on mobile. User can swipe to up to dismiss the notification.
* This component will act as an interface that will manage notifications. It will have two methods to add and remove notifications.

### API Reference
Notification component acts as an interface that will manage notifications. It has two methods to add and remove notifications.

#### Attributes

| Attribute                  | Description                                                             | Default Value |
| -------------------------- | ----------------------------------------------------------------------- | ------------- |
| `duration` (`number`)      | Sets the default duration of notifications in seconds when not provided | `7`           |
| `no-animation` (`boolean`) | Disables animation of notifications                                     | `false`       |

#### Methods

| Method               | Description                                                                                         | Parameters               | Return Value                                           |
| -------------------- | --------------------------------------------------------------------------------------------------- | ------------------------ | ------------------------------------------------------ |
| `addNotification`    | Adds a notification to the notification list                                                        | BlNotificationCard Props | The added notification object with remove method       |
| `removeNotification` | Async method that removes a notification from the notification list after the animation is finished | Notification ID          | A promise that resolves when the animation is finished |

## `bl-notification-card` Implementation
General usage example:

```html
[...]

PR #300, opened 2022-10-19
feat: notification component

onursert:
Closes https://github.com/Trendyol/baklava/issues/141.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -18,7 +18,8 @@ module.exports = {
         'checkbox',
         'alert',
         'select',
-        'pagination'
+        'pagination',
+        'toast'
       ],
     ],
   },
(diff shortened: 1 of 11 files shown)


leventozen:
Hi Onur, thank you for your contribution. However, before coding phase, there should be an issue to work and talk on it. We create a technical document or make a decision inside the issue, then coding phase is started.
We have more or less four different types of issues in Baklava.
1 - Fully technical improvement Issues
2 - Fully design related issues (ux/ui)
3 - Component development Issues
4 - Bugs

In your PR, there are lots of questions should be answered because of the nature of the design system.
- Do we need those features? (/w design team)
- If we need, what will be the experience, what is the right way to add those in terms of user experience? (/w design team)
- What is the right way of develop these features in terms of development? (technical discussion in the issue)
etc.

batuhantozun:
Good news, our notification component status is in design. You can check all decisions and details at #141. Maybe this PR can be start point for us when we starting development 🚀

muratcorlu:
We can re-open this PR once it's ready.

PR #301, opened 2022-10-20
feat: radio component

### LINKED ISSUE #135: Radio Input Component
## Design

[Figma Design Document
](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=1510%3A6857)

## Implementation

General usage example:

```html 
<bl-radio-group label="Payment Type" name="payment-type" value="cc">
    <bl-radio value="cc">Credit Card</bl-radio>
    <bl-radio value="ideal">iDeal</bl-radio>    
</bl-radio-group>
```

### `bl-radio-group` component

This component is the wrapper component of the radio options. Field value will be set/red via this component.

#### Attributes

| --------- | ----------- | ------------- |
| label (`string`) | Label of the form field. This is needed for accessibility needs. | - |
| name (`string`) | Name of the form field | - |
| value (`string`) | Default value of the input | - |
| required (`boolean`) | Sets form field required | `false` |

#### Events

| Event | Description | Event Data |
| ----- | ----------- | ---------- |
| `bl-radio-change` | Will be triggered once radio selection changed | new value as string |

#### CSS Custom Properties

| Property | Description | Default Value |
| -------- | ----------- | ------------- |
| `--bl-radio-direction` | Can be used for showing radio options as columns instead of rows. Options are `row` or `column` | `row` |

### `bl-radio` component

This component defines options for the radio input.

#### Attributes

| value (`string`) | Value of the option | - |
| disabled (`boolean`) | Sets option disabled | `false` |

#### Slots

| Name | Description | Default Content |
| `default` slot | Label of the option | - |

#### Events

| Event | Description | Event Data |
| ----- | ----------- | ---------- |
| `bl-checked` | Will be triggered once this option is selected | - |
| `bl-focus` | Will be triggered once this option is focused | - |
| `bl-blur` | Will be triggered once this option is blurred | - |

muratcorlu:
While working on unit tests, I wanted to share my work on radio input component.

We decided to use `bl-radio` and `bl-radio-option` components in #135 but while on development, I noticed that a grouping component will be also needed for `checkbox` component and `bl-checkbox-option` inside a `bl-checkbox` component didn't look a good naming to me. Because `bl-checkbox` can be used by itself without grouping a list of option (like toggling a single value). So for that kind of cases `bl-checkbox` name would be a simpler usage (like we already have now).

If you also find logical, I'll update issue definition regarding to this and complete this PR by finishing missing parts.

Closes #135


src/components/radio-group/bl-radio-group.css
```diff
@@ -0,0 +1,21 @@
...
+
+legend {
+  font: var(--bl-font-title-3-medium);
+  color: var(--bl-color-content-primary);
+}
+
+.options {
+  display: flex;
+  flex-flow: var(--bl-radio-direction, column) wrap;
+  gap: 16px;
```
olkeoguz, line 19: A variable set here ?

src/components/radio-group/radio/bl-radio.ts
```diff
@@ -0,0 +1,138 @@
...
+  @property()
+  name: string;
+
+  @property()
+  value: string;
+
+  /**
+   * Sets option as disabled
+   */
+  @property({ type: Boolean, reflect: true })
```
olkeoguz, line 33: The selected and disabled bl-radio does not match with the one in figma

Storybook:
image

Figma:
image

Is this on purpose @muratcorlu ?

muratcorlu: Hmm, good catch. I'll fix and add a story for this case.

src/components/radio-group/bl-radio-group.css
muratcorlu: What do you mean? User can set `--bl-radio-direction` but it's `column` by default.

olkeoguz: Sorry, I meant a variable can be set for the gap:16px

muratcorlu: I mean, it could but in design it was 16px and couldn't be sure if we would need an option for that. I think we can add it easily in the future if we see a need.

olkeoguz (approved):

buseselvi (approved):

muratcorlu:
There is a problem about interaction tests in Chromatic. I created [a separate issue](#322) for that. I'll merge this component without focus state screenshot test(@buseselvi tested it on storybook manually, it works anyway) to not block this component for a long time.

PR #313, opened 2022-11-07
fix: pagination items per page options

### LINKED ISSUE #284: bl-pagination's items per page options are not customizable
Currently, the items per page options for the bl-pagination are statically set inside the component.

image

These options need to be provided by the developer as they may not meet their expectations.

The possible solutions could be:
* to take the select component as slot inside the bl-pagination.
* or these options can be attributes.

olkeoguz:
Improvements:
* The "itemsPerPageOptions" array property is added,
* Default items-per-page is changed from 100 to 10,
* The "option-text" attribute is removed since the itemsPerPageOptions property includes texts and values,

The usage:
```
 <bl-pagination current-page="1" total-items="1500" 
</bl-pagination>

    pagination.itemsPerPageOptions = [
         {
        text: '100 Items',
        value: 100,
      },
      {
        text: '250 Items',
        value: 250,
      },
      {
        text: '500 Items',
        value: 500,
      },
      {
        text: '1000 Items',
        value: 1000,
      },
    ]

```

* Lastly, the component had no responsivity before, so little style changes are added.
 This part can be improved if @buseselvi has any feedbacks.

image

Closes #284.

### DIFF
--- a/src/components/pagination/bl-pagination.ts
+++ b/src/components/pagination/bl-pagination.ts
@@ -12,25 +12,6 @@ import style from './bl-pagination.css';
  * @summary Baklava Pagination component
  */
 
-const selectOptions = [
-  {
-    text: '10 Items',
-    value: 10,
-  },
-  {
-    text: '25 Items',
-    value: 25,
-  },
-  {
-    text: '50 Items',
-    value: 50,
-  },
-  {
-    text: '100 Items',
-    value: 100,
-  },
-];
-
 @customElement('bl-pagination')
 export default class BlPagination extends LitElement {
   static get styles(): CSSResultGroup {
@@ -53,7 +34,7 @@ export default class BlPagination extends LitElement {
    * Sets the number of items per page
    */
   @property({ attribute: 'items-per-page', type: Number, reflect: true })
-  itemsPerPage = 100;
+  itemsPerPage = 10;
 
   /**
    * Adds jumper element if provided as true
@@ -80,10 +61,27 @@ export default class BlPagination extends LitElement {
   selectLabel = 'Show';
 
   /**
-   *  Sets the items per page options
+   *  Sets the items per page options of the select element
    */
   @property({ attribute: 'select-options', type: Array })
-  selectOptions = selectOptions;
+  selectOptions = [
+    {
+      text: '10 Items',
+      value: 10,
+    },
+    {
+      text: '25 Items',
+      value: 25,
+    },
+    {
+      text: '50 Items',
+      value: 50,
+    },
+    {
+      text: '100 Items',
+      value: 100,
+    },
+  ];
 
   @state() private pages: Array<number | string> = [];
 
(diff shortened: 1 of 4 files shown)


src/components/pagination/bl-pagination.ts
```diff
@@ -80,10 +61,27 @@ export default class BlPagination extends LitElement {
   selectLabel = 'Show';
 
   /**
-   *  Sets the option texts.
+   *  Sets the items per page options of the select element
    */
-  @property({ attribute: 'option-text', type: String })
-  optionText = 'Items';
+  @property({ attribute: 'select-options', type: Array })
```
muratcorlu, line 66: I don't think it's a good idea to get serialized objects via attributes. I would prefer to get this via a property only (so it can be only set by JavaScript) or with a slot.

Passing stringified json as a property value can cause issues on FW implementations as well.

Can we reconsider this? Let's also see what others think about this.

muratcorlu: Another point here is the property/attribute name. `select-options` doesn't give correct feeling about what we are setting here. `select` is a representational choice for choosing an "item per page" value but, it would be a different element. To keep that decision internal to the component, I would suggest to use a name directly related with the content. Like `itemsPerPageOptions` maybe.

olkeoguz: I'm converting the pr to craft until we decide then. This is the first time we take an array property, so I was in the middle but now I think what you suggest makes total sense @muratcorlu . 
@leventozen  what do you think?

leventozen: I agree with @muratcorlu about his concerns.

olkeoguz:
> We updated the design documentation by considering responsivity. Here is the link: https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=1177%3A9404
> 
> If there is no comment on this, can we update the responsivity part 🙏🏻 @olkeoguz

I set the mobile width to 'below 768px' here. Perhaps we can decide to pick a mobile width that will be same for all the baklava components sometime.

src/components/pagination/bl-pagination.ts
```diff
@@ -80,10 +61,28 @@ export default class BlPagination extends LitElement {
   selectLabel = 'Show';
 
   /**
-   *  Sets the option texts.
+   * Sets the items per page options of the select element
+   *  PROPERTY
    */
-  @property({ attribute: 'option-text', type: String })
-  optionText = 'Items';
+  @property({ type: Array })
```
muratcorlu, line 67: Use `attribute: false` to use this as a property only. Otherwise still component will try to parse this as an attribute.

muratcorlu, line 67:
```suggestion
  @property({ type: Array, attribute: false })
```

Use `attribute: false` to use this as a property only. Otherwise still component will try to parse this as an attribute.

muratcorlu (changes requested):

buseselvi (approved):

Issue #315, opened 2022-11-08
Dialog element is not supported in older versions of browsers


DamlaDemir:
**Describe the bug**
The `html dialog` element is not supported in older versions of browsers. We have users using versions below Safari 15.3 and they see the dialog element printed directly on the screen.

You can check browser supports for html dialog -> https://caniuse.com/dialog

**To Reproduce**
1. Go to https://trendyol.github.io/baklava/?path=/docs/components-dialog--basic-usage in old versions of browsers
2. See error

**Expected behavior**
Dialog element should be supported in older browsers.

We need to add a polyfill for the dialog. When I searched, I came across [this](https://github.com/GoogleChrome/dialog-polyfill) package developed by google chrome. Maybe we can use that, but we'll have to include it in the baklava.

**Screenshots**
![bs_macbsr_Safari_14 1 (1)](https://user-images.githubusercontent.com/26359615/200568381-5be83b58-f529-46d9-b5bd-bf6b8538ead9.jpg)

**Desktop:**
 - OS: MacOs
 - Browser: Safari 
 - Version: 14.3

PR #319, opened 2022-11-16
fix: support of dialog in old versions of browsers

DamlaDemir:
Native html dialog tag is not supported in old versions of browser due to this reason we decided to do new dialog component for unsupported browsers.

We did not add a new component for the `dialog-polyfill` because there was too much code duplication. By adding some controls, we ensured to render the component we wrote with 'div' tags in browsers where the dialog is not supported.
If html dialog is supported in the future, we can easily remove the controls we added.

Closes https://github.com/Trendyol/baklava/issues/315


src/components/dialog/bl-dialog.css
```diff
@@ -1,36 +1,52 @@
...
+.dialog-polyfill {
+  display: var(--display);
+  position: fixed;
+  left: 0;
+  top: 0;
+  min-width: 100vw;
+  min-height: 100vh;
+  align-items: center;
+  justify-content: center;
+  background: #273142b3;
```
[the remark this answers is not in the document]
muratcorlu: Ah I see, we are not able to set alpha with a css variable. Maybe we can consider to add a FIXME here to use CSS variables for alpha colors.

src/components/dialog/bl-dialog.test.ts
```diff
@@ -1,14 +1,8 @@
-import {
-  assert,
-  fixture,
-  html,
-  oneEvent,
-  expect,
-  fixtureCleanup,
-} from '@open-wc/testing';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
```
muratcorlu, line 2: Why do we need those?

src/components/dialog/bl-dialog.ts
```diff
@@ -124,17 +142,13 @@ export default class BlDialog extends LitElement {
...
+    if (window.innerWidth <= 471) {
+      dialogHeight = 128; // 56px(header) + 32px(dialog-margin) + 40px(content-padding)
     }
 
-    this.content.style.maxHeight = `${window.innerHeight - (contentHeight + footerHeight)}px`;
-  }
-
-  get _hasFooter() {
-    return [...this.childNodes].some(node => node.nodeName === 'BL-BUTTON');
+    this.content.style.maxHeight = `${window.innerHeight - (dialogHeight + footerHeight)}px`;
```
muratcorlu, line 151: What are we doing here? Why we are not setting this inside CSS?

src/components/dialog/bl-dialog.ts
```diff
@@ -147,26 +161,32 @@ export default class BlDialog extends LitElement {
...
+      ${this.renderFooter()}
+    </div>`;
+  }
+
+  render(): TemplateResult {
+    return this.hasHtmlDialogSupport
+      ? html`
+          <dialog class="dialog" aria-labelledby="dialog-caption">${this.renderContainer()}</dialog>
+        `
+      : html`<div class="dialog-polyfill" aria-labelledby="dialog-caption">
```
muratcorlu, line 187: I think we also need `aria-role="dialog"` here.

src/components/dialog/bl-dialog.ts
```diff
@@ -77,18 +82,31 @@ export default class BlDialog extends LitElement {
...
+        ? this.dialog.showModal?.()
+        : this.dialogPolyfill.style.setProperty('--display', 'flex');
+        
       this.onOpen({ isOpen: true });
       document.body.style.overflow = 'hidden';
     } else {
-      this.dialog.close?.();
+      this.hasHtmlDialogSupport
+        ? this.dialog.close?.()
+        : this.dialogPolyfill.style.setProperty('--display', 'none');
```
muratcorlu, line 108: We should not need to do this. Let's just use CSS:

```css
:host() .dialog-polyfill {
   display: none;
   ...
}

:host([open]) .dialog-polyfill {
  display: flex;
}
```

src/components/dialog/bl-dialog.ts
```diff
@@ -32,9 +32,12 @@ export default class BlDialog extends LitElement {
   @property({ type: String })
   caption?: string;
 
-  @query('dialog')
+  @query('.dialog')
   dialog: HTMLDialogElement & DialogElement;
 
+  @query('.dialog-polyfill')
+  dialogPolyfill: HTMLDivElement;
```
muratcorlu, line 39: Set this private.

src/components/dialog/bl-dialog.ts
```diff
@@ -32,9 +32,12 @@ export default class BlDialog extends LitElement {
   @property({ type: String })
   caption?: string;
 
-  @query('dialog')
+  @query('.dialog')
   dialog: HTMLDialogElement & DialogElement;
```
muratcorlu, line 36: Set this private.

src/components/dialog/bl-dialog.css
```diff
@@ -1,36 +1,52 @@
...
+.dialog::backdrop {
   background: #273142;
   opacity: 0.7;
 }
 
-dialog .container {
-  min-width: 424px;
-  min-height: 178px;
+.dialog-polyfill {
+  display: var(--display);
```
muratcorlu, line 20: Use `[open]` selector instead of variable. Check below for the detailed comment.

src/components/dialog/bl-dialog.css
```diff
@@ -1,36 +1,52 @@
...
+  position: fixed;
+  left: 0;
+  top: 0;
+  min-width: 100vw;
+  min-height: 100vh;
```
muratcorlu, line 25: I think we should just set width and height, not minimum values. Dialog background should not be scrollable, right?

src/components/dialog/bl-dialog.ts
DamlaDemir: Good idea  🎉

src/components/dialog/bl-dialog.css
DamlaDemir: Yes you are right, setting min-max value is enough for this. I actually fixed it here, but it may have come back with ctrl+z 🥲

src/components/dialog/bl-dialog.test.ts
DamlaDemir: We wanted to mock `window.HTMLDialogElement` object in tests as true or false value but typescript don't let this operation. If you have any idea for this it would be very help .

buseselvi (approved):

olkeoguz (approved):

src/components/dialog/bl-dialog.css
```diff
@@ -1,36 +1,62 @@
...
+  display: none;
+  position: fixed;
+  left: 0;
+  top: 0;
+  width: 100vw;
+  height: 100vh;
+  align-items: center;
+  justify-content: center;
+  background: #273142b3; /* FIX ME -> Use css variables for alpha colors */
+  z-index: 999; /* FIX ME -> Use z-index variable */
```
muratcorlu, line 40:
```suggestion
  /* FIXME: Use css variables for alpha colors */
  background: #273142b3;
  /* FIXME: Use z-index variable */
  z-index: 999;
```

playground/template.html
```diff
@@ -29,5 +29,189 @@ <h1>Baklava Playground</h1>
...
+  });
+
+  dialog2.addEventListener('bl-dialog-open', d => {
+    console.log(d);
+  });
+
+  dialog2.addEventListener('bl-dialog-close', d => {
+    console.log(d);
+  });
+</script>
```
muratcorlu, line 217: Please don't change template for component specific works 😊

src/components/dialog/bl-dialog.css
```diff
@@ -1,36 +1,62 @@
...
+.dialog,
+.dialog-polyfill .container {
+  padding: 0;
+  border: 0;
+  border-radius: var(--bl-border-radius-l);
+}
+
+.dialog-polyfill .container {
+  position: fixed;
+  z-index: 999; /* FIX ME -> Use z-index variable */
```
muratcorlu, line 22:
```suggestion
  /* FIXME: Use z-index variable */
  z-index: 999;
```

muratcorlu (changes requested):

src/components/dialog/bl-dialog.css
DamlaDemir: Is this a general rule ? The linter expects empty line before comment. So this looks like this:


image

muratcorlu: Yeah, I think it's for making it more noticeable. But still, I think it should also be like `FIXME:` instead of `FIX ME ->`

muratcorlu (approved):

Issue #322, opened 2022-11-17
Visual regression test for focus state of Radio Input


muratcorlu:
In the [related PR](#301) I couldn't manage to run interaction tests on chromatic to get focus state screenshots. I contacted with Chromatic support and got an reply below:

> Hey Murat!
>Thanks for your patience - I wanted to provide you with a quick update regarding the issue you reported around interaction tests. We've identified that the issue is a bug, and our Engineering team is looking into what's causing it. I'll update you once a resolution has been reached - let me know if you have any questions in the meantime!
> Regards,
> Suki

Now we are waiting for a fix on Chromatic, then we'll need to check radio input focus state screenshots.

muratcorlu:
We had an answer:

> Hey Murat - thanks for your patience!
> Our Engineers determined the reason your interaction tests are having issues showing is due to the Storybook version you're using. If you upgrade your Storybook version to 6.5.10 or later, you should see the interactions badge appear. Let me know if you have any questions or if you still experience issues after upgrading!
> Regards,
> Suki

I'll upgrade Storybook with this issue and check if it really fix the problem.

PR #324, opened 2022-11-23
feat: checkbox-group component

### LINKED ISSUE #290: Accessibility improvements on Checkbox component
[Figma](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=606%3A4458&t=EeXAS4ruHbwtJQAN-1)
- [ ] Checkbox component should be usable with keyboard navigation.
- [ ] Checkbox component should use related `aria-` attributes. [More info](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role)

### LINKED ISSUE #291: Using checkbox component in form
The checkbox component should be usable as a form input by implementing ElementInternals. We may also consider having a grouping component for using a checkbox as a form field. Like `checkbox-group` or `bl-fieldset` etc.

olkeoguz:
Includes:
 * checkbox-group component of which the implementation was discussed in #291:
```html
 <bl-checkbox-group id="sportSelection" label="Choose sports you like" name="sports" required value='["basketball","tennis"]'>
  <bl-checkbox value="football">Football</bl-checkbox>
  <bl-checkbox value="basketball">Basketball</bl-checkbox>
  <bl-checkbox value="tennis">Tennis</bl-checkbox>
</bl-checkbox-group>
```

```js
<script>
  sportSelection.value; // ['basketball', 'tennis']
</script>
```

 * accessibility and form control improvements of checkbox component mentioned in #290, 
 
checkbox focus state screenshots will be added after #322 resolves.

Closes #290, closes #291.


src/components/checkbox-group/bl-checkbox-group.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+   * Sets option as required
+   */
+  @property({ type: Boolean, reflect: true })
+  required = false;
+
+  get options(): BlCheckbox[] {
+    return [].slice.call(this.querySelectorAll(blCheckboxTag));
+  }
```
muratcorlu, line 45: You may prefer to use `@queryAll(blCheckboxTag)` of lit.

muratcorlu (approved):
Looks very good! I just have a small remark. And let's also ask for a design review from @buseselvi 😊

src/components/checkbox-group/bl-checkbox-group.ts
olkeoguz: I tried and it didn't select checkboxes somehow. Could @queryAll selector not support selecting the elements in shadowRoot ?

muratcorlu: Main purpose of it is querying the elements inside the component shadowRoot. It basically does what you do here:

> A property decorator that converts a class property into a getter that executes a querySelectorAll on the element's renderRoot.
> https://lit.dev/docs/api/decorators/#queryAll

But it doesn't return an array of `BlCheckbox` It returns `NodeListOf<BlCheckbox>`. Maybe that could be the issue?

olkeoguz:
```
 @queryAll(blCheckboxTag) checkboxes: NodeListOf<BlCheckbox>;

  get options(): BlCheckbox[] {
    console.log("checkboxes",this.checkboxes);
    return [].slice.call(this.checkboxes);
  }
```

image

Is there something missing ?

olkeoguz:
After searching and discussing checkbox-group examples with @buseselvi, we decided to change the keyboard navigation. Pressing tab will always switch to the next element in the group rather than focusing outward. 
Example : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

But navigating with arrow keys will be the same as now. 
I'll make these changes shortly if that's final @muratcorlu  @buseselvi .

buseselvi (approved):

src/components/checkbox-group/bl-checkbox-group.ts
olkeoguz: Could you give it a shot on your local @muratcorlu  ? I couldn't debug the problem and make it work properly.

muratcorlu: `@query` and `@queryAll` are filled after `firstUpdated` hook. When do you call this getter? Can it be before `firstUpdated`?

olkeoguz: Yes, I call it right after mount. I need it to check all the available or checked checkboxes. Just like we did in radio-group actually.

PR #333, opened 2022-12-06
feat: dropdown component

### LINKED ISSUE #239: Dropdown Component
[Figma](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=2158%3A8508)

leventozen:
Closes #239.


src/components/dropdown/group/bl-dropdown-group.css
```diff
@@ -0,0 +1,38 @@
...
+}
+
+/* adding seperator before and after of element, except not first child and last child of the dropdown */
+:host(:not(:first-child))::before,
+:host(:not(:last-child))::after {
+    position: absolute;
+    content: '';
+    width: 100%;
+    border-bottom: 1px solid var(--bl-color-border);
+}
```
muratcorlu, line 26: `border-bottom` and `border-top` would be enough here, I think. Like

```suggestion
:host(:not(:first-child))::before {
    border-bottom: 1px solid var(--bl-color-border);
}

:host(:not(:last-child))::after {
    border-top: 1px solid var(--bl-color-border);
}
```

src/components/dropdown/item/bl-dropdown-item.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    if (!this.BlDropdownField && !this.BlDropdownGroupField) {
+      console.warn(`bl-dropdown-item is designed to be used inside a ${blDropdownGroupTag} or ${blDropdownTag}`, this);
+    }
+
+  }
+
+  disconnectedCallback(): void {
+    super.disconnectedCallback();
+  }
```
muratcorlu, line 57: We don't need this if we'll not do anything custom here.

src/components/dropdown/item/bl-dropdown-item.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  render(): TemplateResult {
+    return html` <bl-button
+      variant="tertiary"
+      kind="neutral"
+      icon="${ifDefined(this.icon)}"
+      @click="${this._handleClick}"
```
muratcorlu, line 64: Maybe it's better to listen our own `bl-click` event here. 🤔

src/components/dropdown/bl-dropdown.ts
```diff
@@ -0,0 +1,170 @@
...
+   */
+  @event('bl-dropdown-close') private onClose: EventDispatcher<string>;
+
+  connectedCallback() {
+    super.connectedCallback();
+  }
+  disconnectedCallback() {
+    super.disconnectedCallback();
+
+    this._cleanUpPopover && this._cleanUpPopover();
```
muratcorlu, line 90: Do `document.removeEventListener('click', this._handleClickOutside);` if dropdown is open here.

src/components/button/bl-button.ts
```diff
@@ -105,14 +117,28 @@ export default class BlButton extends LitElement {
     });
   }
 
+  get _isActive() {
+    if (!this.dropdown) {
+      return false
+    }
```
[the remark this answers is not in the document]
leventozen: I agree with you. I am working on it 💯

src/components/dropdown/bl-dropdown.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  get isPopoverOpen() {
```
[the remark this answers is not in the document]
leventozen: Actually, I copied the way of naming these popovers from another components that we built however I agree with you. I will update those as well.

src/components/dropdown/item/bl-dropdown-item.ts
leventozen: Adding disconnectedCallback with connectedCallback seems more clear and natural to me when I look at the code. However, I agree with you, there is no need. What do you think?

src/components/dropdown/group/bl-dropdown-group.css
leventozen: 🤔 It wouldn't work without those.
```css
position: absolute;
content: '';
width: 100%;
```
However `::after` and `::before` is not necessary here. I will delete those too to simplify like you suggested.

src/components/dropdown/item/bl-dropdown-item.ts
leventozen: I changed this to `@bl-click`.
When I made this change, we must trigger button element inside the `shadowRoot` rather than the host element. We may need a strategy that why/how we create custom events etc. Thanks for the suggestion.

src/components/dropdown/bl-dropdown.css
```diff
@@ -0,0 +1,29 @@
...
+    display: inline-block;
+}
+
+.popover {
+    --left: 0;
+    --top: 0;
+
+    position: fixed;
+    z-index: 1;
+    display: var(--bl-dropdown-display, none);
```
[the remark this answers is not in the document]
leventozen: Good one, I deleted the variable.

src/components/dropdown/bl-dropdown.ts
leventozen: I don't think we need this. We already have this when user calls `open()`.

muratcorlu: If a dropdown is removed from DOM while it's "open", click event listener will stay in memory. I suggested this to prevent that situation.

src/components/dropdown/group/bl-dropdown-group.css
muratcorlu: I just felt our aim is as same as adding a border-top or border-bottom to the group element here, isn't it?

Without `::after` and `::before`, `content` will not do anything.

leventozen:
> Good job! I have some remarks. Also:
> 
> * Button component now has a new property, a state and a new CSS variable. Let's also document them in Storybook with some demonstrations.
> * Please check if dropdown is usable via keyboard navigation
> * Please add proper ARIA attributes. `menu`, `group` and `menuitem` roles would be relevant I believe. Also don't miss `aria-labelledby` for group caption.

* Actually, I think `dropdown` property should be private for the user because it is only useful in `bl-dropdown`. However, documenting the active state and css variable is a good idea.
* Forgot to add, I am working on the rest. Thanks

src/components/dropdown/bl-dropdown.css
```diff
@@ -0,0 +1,29 @@
...
+    flex-direction: column;
+    align-items: flex-start;
+    padding: var(--bl-size-m);
+    gap: var(--bl-size-xs);
+    overflow-y: auto;
+    background: var(--bl-color-primary-background);
+    border: 1px solid var(--bl-color-border);
+    box-shadow: 0 10px 15px -8px #27314226;
```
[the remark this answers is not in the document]
leventozen: We can add `FIXME` there. Since we don't have any variable or variable scheme that solve this.

src/components/dropdown/bl-dropdown.ts
leventozen: Is there a way to produce this? TBH I don't understand what removing means.

muratcorlu: In web components `connected` means a copy of this tag is just added to document. And [`disconnected`](https://lit.dev/docs/components/lifecycle/#disconnectedcallback) means it’s removed from DOM. Let’s consider you are using this component in VueJS:

```html
<bl-dropdown v-if=“isAdmin”> …. </bl-dropdown>
```

It’s possible that `isAdmin` can change regarding to another action in the page. And if it switches to `false` when user opened that dropdown, then it’ll be removed from DOM. Currently when you open a dropdown you start listening `click` event of the document. And normally you remove this listener when you close the dropdown but in that case, close couldn’t be happened since it’s completely removed from document. Then that event listener stays in the memory.

src/components/dropdown/group/bl-dropdown-group.css
leventozen: Yeah, we are completely in the same page here.

buseselvi:
Looks great! The only thing is that the popover borders should take the color of the buttons for each kind, I updated the [design file](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=2158%3A8508&t=oNOIhxL3xLoLguEx-1) according to that.  And the popovers caption font should be --bl-font-caption 🚀 @leventozen

buseselvi (approved):

muratcorlu (approved):

Issue #265, opened 2022-10-03
Adding search feature to Select Component


batuhantozun:
[Figma](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=2562%3A21640)

We need to add searchable feature to our select component.
Details will be here...

leventozen:
## Search Feature in Select Component Design Meeting Report

Date: 26.10.2022 15:30 (UTC+3)

* We don't need to a have a `Search component` for this feature.
* We'll need an enhancement on the input component to have a "search input" version (with a search icon at left and clear icon at right to clear search inputs.)
* Search input will be sticky.
* When select opened, search input will be focused.
* When I type, I shouldn't see selected items. Otherwise, those should be at top.

Design work will start according to those decisions then implementation will be planned.

pyilmazer:
Implementation details question : Will there be any count limit to separate in memory items flow from server side processing flow? This behaviour may be determined with a prop. Ex: Apply in memory items flow when item count is less than 100.

batuhantozun:
We need to decide some details regarding to our update meeting. Here is the details;

- Adding loading state to input component
- Adding no data status to search results
- Can we move search icon to right of input component?

@buseselvi FYI

musaakkayaa:
## Design

[Figma](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=25%3A3606&t=2nNj1LncV2RG17sa-0)

## Implementation

General usage example:

```html 
<bl-select>
    <bl-select-option value="tr">Turkey</bl-select-option>
    <bl-select-option value="nl">Netherlands</bl-select-option>
</bl-select>
```

### Rules

## API Reference:

### `bl-select` component

Description

#### Attributes

| Attribute | Description | Default Value |
| --------------- | --------------- | --------------- |
| label (`string`) | Sets the label value (optional) | - |
| placeholder(`string`) | Sets the placeholder value. If left blank, the label value (if specified) is set as placeholder. (optional) | - |
| size(`string`) | Sets the size value. Select component's height value will be changed accordingly (large, medium, small) | medium |
| required (`boolean`) | When option is not selected, shows component in error state | false |
| disabled (`boolean`) | Shows the component in disabled state | false |
| multiple (`boolean`) | Allows multiple options to be selected | false |
| label-fixed (`boolean`) | Makes label as fixed positioned | false |
[...]

Issue #192, opened 2022-08-08
Drawer Component


muratcorlu:
## Design

[Figma](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=2963%3A14146&t=EeXAS4ruHbwtJQAN-1)

## Implementation

General usage example:

```html 
<bl-drawer caption='Drawer Title' external-link='https://some-url.com'>
    content
</bl-drawer>
```

### Rules

* Title and external link displayed on header section in drawer. Title is necessary but external link is optional.
* Close button always displayed on header section and drawer can be closed by clicking close button.
* By default, Drawer can not close by clicking somewhere outside drawer.
* Drawer appears right side on the page with full height expanded.
* Title can be multiline automatically if it does not fit one line. 
* If user does not specify title, title part of header will be rendered empty.
* There will be an animation when drawer opens and closes. ( Animation topics will be discussed again in later times )
* There is an attribute about iframe and drawer component handle iframe rendering itself.
* Only one drawer can display at the same time. When one drawer opens others will be closed.

### `bl-drawer` component

Description

#### Attributes

| Attribute | Description | Default Value |
| --------------- | --------------- | --------------- |
| caption (`string`) | Title of the drawer (optional) | - |
| external-link (`string`) | External link (optional) | - |
| embed-url (`string`) | Iframe url (optional) | - |
| open (`boolean`) | Sets drawer open-close status | - |

#### Slots

| Name | Description | Default Content |
| --------------- | --------------- | --------------- |
| `default` slot | Content inside drawer  | - |

#### Events

| Event | Description |
| --------------- | --------------- |
| `bl-drawer-open` | Will be triggered every time drawer is opened |
| `bl-drawer-close` | Will be triggered every time drawer is closed |

muratcorlu:
# Drawer Component Design Meeting Report

Date: 16.11.2022 15:30 (UTC+3)

* There will be no backdrop (like modal dialog) in first phase.
* Title would be set as a parameter and title & close button will be sticky on the panel
* Content area will be a free area that anyone can be put
* We may consider some technical helpers for the case to put an iframe as drawer content. We'll check potential issues for putting an iframe inside drawer (like loading iframe even when drawer is closed)
* When a drawer is visible and user opens another drawer, new drawer will come over the current one. And they will be closed one-by-one. (Of course app developer can decide to close other or all in a programmatic way)
[...]

PR #336, opened 2022-12-07
feat: added search input for select component

musaakkayaa:
## Design

[Figma](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=25%3A3606&t=2nNj1LncV2RG17sa-0)

## Implementation

General usage example:

```html 
<bl-select>
    <bl-select-option value="tr">Turkey</bl-select-option>
    <bl-select-option value="nl">Netherlands</bl-select-option>
</bl-select>
```

### `bl-select` component

Description

#### Attributes

| label (`string`) | Sets the label value (optional) | - |
| placeholder(`string`) | Sets the placeholder value. If left blank, the label value (if specified) is set as placeholder. (optional) | - |
| size(`string`) | Sets the size value. Select component's height value will be changed accordingly (large, medium, small) | medium |
| required (`boolean`) | When option is not selected, shows component in error state | false |
| disabled (`boolean`) | Shows the component in disabled state | false |
| multiple (`boolean`) | Allows multiple options to be selected | false |
| label-fixed (`boolean`) | Makes label as fixed positioned | false |
| help-text (`string`) | Adds help text(optional) | - |
| invalid-text (`string`) | Set custom error message(optional) | - |
| search-bar (`boolean`) | Show search text for options(optional) | - |
| search-bar-placeholder (`string`) | Search text placeholder(optional) | - |
| search-not-found-text (`string`) | When options not found, shown this text | Not Found |
| search-bar-loading-state (`boolean`) | Search bar loading state for fetch request | false |
| search-filterable (`boolean`) | Search in the option list | true |

#### Slots

| Name | Description | Default Content |
| `default` slot | Content inside select  | - |
| `search-not-found` slot | Custom show search not found layout | - |

#### Events

| Event | Description |
| --------------- | --------------- |
| `bl-select` | Will be triggered when click option |
| `bl-search` | Will be triggered every time search bar key down |


muratcorlu:
Hi @musaakkayaa . Thanks for your contribution!

* Please write some description to PR about the work, your decisions and ideas behind those decisions.
* We are not using issue reference numbers in commit messages. Use closing words in PR description instead.
* Storybook stories don’t cover all of the options we provide.
* It would be great to first share your technical solution proposal in issue (#265) like we do in some other issues (#192).

I’m not sure if the current way of doing search (searching with the text content of dom elements) is the best approach for the needs. Generally this search will be added if we have so much options, like hundreds of items. Currently, we always need to provide all options in DOM to be able to use search functionality. Also it doesn’t look like possible to use search UI for searching via an external API. I don’t think a search functionality like this will be helpful for most of the cases.

I’m curious about your and others’ comments about this.

muratcorlu:
Hi @musaakkayaa I'm closing this PR for now to conclude ADR discussion first. You can re-open once you think it's ready.

Issue #330, opened 2022-12-02
Using an example external iframe example for Drawer component for Storybook


muratcorlu:
[Current Drawer Component implementation](#327) doesn’t have a proper storybook example of embedded external content (iframe) variant.

- [ ] An iframe URL should be used that doesn’t have an ever-changing content (At least content changing should be under control)
- [ ] Iframe URL should load quickly and not block Visual Regression Tests on Chromatic

We may consider to include a test HTML file in codebase to deploy with Storybook and show it inside drawer.

PR #345, opened 2022-12-22
fix: remove fouce effect styles from default theme

### LINKED ISSUE #344: Web components don't appear on hotjar records
**Describe the bug**
We are implementing below styles to avoid the FOUCE effect but this causes web components not appearing in hotjar recordings.
```css
/* Prevent FOUCE */
:not(:defined) {
  visibility: hidden;
}
```

**Screenshots**
We have bl-tab-group, bl-button, bl-badge and etc. that don't appear as below.
image

**Additional context**
We should let the developers deal with this issue and apply these styles in their own applications.

olkeoguz:
Includes: 
* Removing FOUCE effect style,
* Adding FOUCE effect warning to docs.

Closes #344.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,4 +23,10 @@ Preferred way of using Baklava is using it via CDN. Just import library JS and C
 
 This way library will be served from a very high performant CDN and all of the Baklava web components will be ready to use inside your web project.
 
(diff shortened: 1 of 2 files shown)


muratcorlu (approved):
Good catch! I was already hesitant about having this style in a css file that we claim as "theme". We may improve documenting this issue later.

muratcorlu:
I see UI Test of Drawer component is failing because of the prioritization of iframe in it. I think we can ignore it for now by delegating fixing this to #330

Issue #207, opened 2022-08-23
Visual regression test for input component states


batuhantozun:

muratcorlu:
Same problem applies also for other form components like textarea and select. I'm adding this issue to our board to give more priority.

PR #347, opened 2022-12-23
feat: textarea component

### LINKED ISSUE #145: Textarea Component
## Design

[Figma Design Document](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=2641%3A11444)
## Implementation

General usage example:

```html 
<bl-textarea value="Example"></bl-textarea>
```

### Rules

Put some rules about how this component should behave and implemented

* Component will have a default width but can be set by the developer (as in input component)
* By default textarea can be resized by user vertically (Using default behavior of native textarea input)
* Visible row count can be set by the developer (with a default value, 4)
* Optionally, textarea can expand by entering text. In this mode, resizing mode will be disabled. A max row count value can be set by the developer (with a default value)
* Visually, textarea should give the feel that it's in scrollable mode (by showing some part of the text above or below)
* Label styles will be same as the Input component
* Helper and error text, placeholder styles functionality will be provided as in input
* Textarea can have minlength, maxlength validation rules
* Optionally, an active character count can be visible
* If textarea maxlength is provided, this value will be visible inside character count (as like 10 / 100)
* Active character count can exceed the maxlength, but in this case textarea will be in invalid state. And if character counter is visible, this issue will be highlighted in active char count.
* Minlength validation will be visible on character count once user finished entering text (with losing focus for the first time, as in input component)

### `bl-textarea` component

Textarea component is the component to take multi-line text input from user.

#### Attributes

| value (`string`) | Value of textarea | - |
| name (`string`) | Name attribute to use textarea inside forms |  - |
| required (`boolean`) | Sets required property of textarea |  - |
| size (`string`) | Sets input size |  medium |
| rows (`number`) | Row count of textarea | 4 |
| expand (`boolean`) | Expands by entering text | false |
| max-row (`number`) | Maximum row count when text area expands | - |
| placeholder (`string`) | Sets the placeholder of textarea | - | 
| label (`string`) | Sets the label of textarea | - | 
| label-fixed (`boolean`) | Makes label as fixed positioned | false | 
| minlength (`number`) | Minimum length of the textarea | - | 
| maxlength (`number`) | Maximum length of the textarea | - | 
| invalid-text (`string`) | Custom error message | - | 
| help-text (`string`) | Helper text of textarea | - | 
| character-count (`boolean`) | Sets visibility of character count | false | 

#### Events

| Event | Description |
| --------------- | --------------- |
| `bl-change` | Fires when an alteration to the element's value is committed by the user. Unlike the bl-input event, the change event is not necessarily fired for each alteration to an element's value. |
| `bl-input` | Fires when the value of an textarea element has been changed. |
| `bl-focus` | Fires when user focuse to textarea.  |
| `bl-valid` | Fires when the value of an input element has been changed. |
| `bl-invalid` | Fires when the value of an input element has been changed. |

ozkersemih:
Closes #145


src/components/textarea/bl-textarea.css
```diff
@@ -0,0 +1,131 @@
+:host {
+  display: inline-block;
+  width: 200px;
+  position: relative;
+
+  --bl-textarea-padding-vertical: var(--bl-size-2xs);
+  --bl-textarea-padding-horizontal: var(--bl-size-xs);
+  --bl-textarea-border-color: var(--bl-color-border);
+  --bl-textarea-icon-color: var(--bl-color-content-tertiary);
+  --bl-textarea-text-color: var(--bl-color-content-primary);
```
muratcorlu, line 10: Those variables should be kept internal. So

- [ ] Move them away from `:host` element since host element can be set outside of shadowdom. You can create a wrapping element if you need.
- [ ] remove `bl-textarea` prefixes.

src/components/textarea/bl-textarea.css
```diff
@@ -0,0 +1,131 @@
...
+  --bl-input-height: var(--bl-size-2xl);
```
muratcorlu, line 11: Do we need this?

src/components/textarea/bl-textarea.ts
```diff
@@ -0,0 +1,235 @@
...
+
+
+  render(): TemplateResult{
+    const maxLengthInvalid = this.internals.validity.rangeOverflow;
+    const invalidMessage = !this.checkValidity()
+      ? html`<p class="invalid-text">${this.validationMessage}</p>`
+      : ``;
+    const helpMessage = this.helpText ? html`<p class="help-text">${this.helpText}</p>` : ``;
+
+    const label = this.label ? html`<label>${this.label}</label>` : '';
```
muratcorlu, line 199: Give an `id` to textarea and put label `for` attribute.

src/components/textarea/bl-textarea.ts
```diff
@@ -0,0 +1,235 @@
...
+      this.reportValidity();
+    });
+  }
+
+  private onKeydown = (event: KeyboardEvent): void => {
+    if (event.code === 'Enter' && this.form) {
+      submit(this.form);
+    }
+  }
+
```
muratcorlu, line 148: Textarea can be multiline. `Enter` key should not submit the form.

src/components/textarea/bl-textarea.ts
```diff
@@ -0,0 +1,235 @@
...
+  maxlength?:number;
+
+  /**
+   * Sets initial value of the textarea
+   */
+  @property()
+  value = '';
+
+  /**
+   * Sets input size.
```
muratcorlu, line 115:
```suggestion
   * Sets textarea visible row count.
```

src/components/textarea/bl-textarea.ts
```diff
@@ -0,0 +1,235 @@
...
+  labelFixed = false;
+
+  /**
+   * Sets placeholder of the textarea
+   */
+  @property({})
+  placeholder?: string;
+
+  /**
+   * Character counter
```
muratcorlu, line 85:
```suggestion
   * Enables showing character counter.
```

src/components/textarea/bl-textarea.css
```diff
@@ -0,0 +1,131 @@
...
+  margin: 0;
+}
+
+.max-len-invalid ~ .brief > .invalid-text {
+  display: inline-block;
+}
+
+.dirty:invalid ~  .brief > .invalid-text {
+  display: inline-block;
+}
```
muratcorlu, line 121:
```suggestion
:where(.max-len-invalid, .dirty:invalid) ~ .brief > .invalid-text {
  display: inline-block;
}
```

src/components/textarea/bl-textarea.css
```diff
@@ -0,0 +1,131 @@
...
+
+.max-len-invalid ~ .brief > .counter-text {
+  color: var(--bl-color-danger);
+}
+
+
+.dirty:invalid ~ .brief > .counter-text {
+  color: var(--bl-color-danger);
+}
```
muratcorlu, line 97:
```suggestion
:where(.max-len-invalid, .dirty:invalid) ~ .brief > .counter-text {
  color: var(--bl-color-danger);
}
```

src/components/textarea/bl-textarea.css
```diff
@@ -0,0 +1,131 @@
...
+  --bl-textarea-border-color: var(--bl-color-border);
+  --bl-textarea-icon-color: var(--bl-color-content-tertiary);
+  --bl-textarea-text-color: var(--bl-color-content-primary);
+  --bl-input-height: var(--bl-size-2xl);
+}
+
+textarea {
+  outline: none;
+  box-sizing: border-box;
+  border: solid 1px var(--bl-textarea-border-color);
```
muratcorlu, line 17: I don't know if we discussed this but component should have a `:focus` or `:focus-within` style for keyboard navigation.

muratcorlu (commented):
I added some early review points.

src/components/textarea/bl-textarea.css
ozkersemih: I removed --bl-textarea.. variables. I use Baklavas own variables.

src/components/textarea/bl-textarea.ts
ozkersemih: I added id that named "bl-text-area". But maybe you have better suggestion abi 😄

muratcorlu: Since this is inside ShadowDOM and there is not risk to have a conflict with other elements outside, you can use whatever you want. You don’t need to use any prefix.

muratcorlu:
Safari currently doesn’t support ElementInternals API. You need to import element-internals-polyfill as we did in input component.

```ts
import 'element-internals-polyfill';
```

ozkersemih:
> 

Thanks thats the solution :)

ozkersemih:
I didnt add size attribute. Because I see widths are same on different size textarea. They have different heights. I think we already control it with rows attribute. In rfc&adr stage I added bl-focus and bl-valid event but we really need them?

src/components/textarea/bl-textarea.ts
```diff
@@ -0,0 +1,277 @@
...
+  valueChangedCallback(value: string): void {
+    this.value = value;
+  }
+
+  validityCallback(): string | void {
+    return this.customInvalidText || this.validationTarget?.validationMessage;
+  }
+
+  private autoResize() {
+    const verticalPadding = parseInt(getComputedStyle(this.validationTarget).padding[0]) * 2;
```
muratcorlu, line 200: `getComputedStyle` is an expensive operation and we do this in every keypress when expand mode is enabled. We may look for an option that we can ignore padding sizes (like using `box-sizing: content-box` in css).

src/components/textarea/bl-textarea.ts
```diff
@@ -0,0 +1,277 @@
...
+    const scrollHeight = this.validationTarget.scrollHeight;
+    if(!this.initialHeight) {
+      this.initialHeight = scrollHeight;
+      this.lineHeight = (this.initialHeight-verticalPadding) / this.rows!;
```
muratcorlu, line 204: If we really need line height, I think we can read it from a variable, instead of trying to calculate it (we set line-height with related font definition in our design system)

muratcorlu (commented):
I still have some doubts about `autoResize` method but I need more time to think about it. In the meantime I added my other remarks.

src/components/textarea/bl-textarea.css
```diff
@@ -0,0 +1,137 @@
...
+  width: 100%;
+  font: var(--bl-font-title-3-regular);
+  padding:var(--bl-size-2xs) var(--bl-size-xs);
+  margin: 0;
+  border-radius: 4px;
+  color: var(--bl-color-content-primary);
+}
+
+textarea:disabled{
+  background-color: var(--bl-color-secondary-background);
```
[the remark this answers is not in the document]
buseselvi: @muratcorlu We applied the same style in the input component. Secondary-background for background and content-passive for text in disabled versions.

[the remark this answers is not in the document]
muratcorlu: But you know, if we use blue for the secondary color, then disabled inputs will be also blue, right? With our color palette issue, we'll already try to address those issues but since this is a new component, maybe we can start to avoid using secondary colors instead of content colors.

buseselvi:
About resize, should we allow to resize its width? Wouldn't it be better if it could just change its height to avoid layout problems like in the image?

image

src/components/textarea/bl-textarea.ts
ozkersemih: if I only call getComputedStyle when component first created( I mean in connectedCallback), is it still expensive for our logic?

src/components/textarea/bl-textarea.test.ts
```diff
@@ -0,0 +1,180 @@
...
+
+      setTimeout(() => textarea?.dispatchEvent(new Event('change')));
+
+      const ev = await oneEvent(el,'bl-change');
+      expect(ev).to.exist;
+      expect(ev.detail).to.be.equal('some value');
+    });
+  });
+  describe('form integration', () => {
+    it('should show errors when parent form is submitted', async () => {
```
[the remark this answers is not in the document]
ozkersemih: Can you explain again? I don't understand clearly

src/components/textarea/bl-textarea.css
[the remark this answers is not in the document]
ozkersemih: So what is our decision about that?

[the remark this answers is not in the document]
buseselvi: we can continue with the colors in the design @ozkersemih

src/components/textarea/bl-textarea.ts
muratcorlu: I you do that calculation in `connectedCallback` that will mean it'll not work properly if that value changes after component rendered (maybe with a CSS media query). So I would not prefer that.

src/components/textarea/bl-textarea.ts
ozkersemih: Is using fixed value for line height not problem for different browser or similar cases?

muratcorlu:
> > About resize, should we allow to resize its width? Wouldn't it be better if it could just change its height to avoid layout problems like in the image?
> > image
> 
> What should we do about that? This behavior is normal textarea behavior I think isn't it? If we want to fix how can we handle it? @muratcorlu

It's possible to customize this. Just use `resize: vertical;` by default. https://developer.mozilla.org/en-US/docs/Web/CSS/resize

src/components/textarea/bl-textarea.ts
ozkersemih: I assigned verticalPadding in firstUpdated() method. So, it called only once when component first updated.

src/components/textarea/bl-textarea.ts
```diff
@@ -0,0 +1,277 @@
...
+   * Sets max row when expand is true
+   */
+  @property({ type: Number, reflect: true, attribute:'max-row' })
+  maxRow?: number;
+
+  /**
+   * Sets textarea size.
+   */
+  @property({ type: String, reflect: true })
+  size?: TextareaSize = 'medium';
```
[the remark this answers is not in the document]
ozkersemih: do we need this? I think textarea doesn't need size values like "medium" "small"

src/components/textarea/bl-textarea.ts
ozkersemih: I think we really need this because of computing current line

src/components/textarea/bl-textarea.ts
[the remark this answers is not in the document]
muratcorlu: Then you can remove it 😊

buseselvi (changes requested):
Character Counter text color should be the same as Helper Text color `--bl-color-content-secondary` by default

buseselvi (approved):

src/utilities/form-control.test.ts
```diff
@@ -1,7 +1,8 @@
 import { elementUpdated, expect, fixture, fixtureCleanup } from "@open-wc/testing";
 import { html, LitElement } from "lit";
 import { customElement, query } from "lit/decorators.js";
-import { innerInputValidators } from "./form-control"
+import {innerInputValidators, textareaLengthValidator} from "./form-control"
+// import BlTextarea from "../components/textarea/bl-textarea";
```
muratcorlu, line 5: Commented code

src/components/textarea/bl-textarea.stories.mdx
```diff
@@ -0,0 +1,216 @@
...
+      invalid-text='${ifDefined(args.customInvalidText)}'
+      help-text='${ifDefined(args.helpText)}'
+      minlength='${ifDefined(args.minlength)}'
+      maxlength='${ifDefined(args.maxlength)}'
+      rows='${ifDefined(args.rows)}'
+      max-row='${ifDefined(args.maxRow)}'
+      expand='${ifDefined(args.expand)}'
+      size='${ifDefined(args.size)}'
+      character-counter='${ifDefined(args.characterCounter)}'
+>${ifDefined(args.child)}</bl-textarea>`
```
muratcorlu, line 62: There can be no slotted content anymore.

src/components/textarea/bl-textarea.ts
```diff
@@ -0,0 +1,256 @@
...
+   * Sets expandity
+   */
+  @property({ type: Boolean, reflect: true })
+  expand = false;
+
+  /**
+   * Sets max row when expand is true
+   */
+  @property({ type: Number, reflect: true, attribute:'max-row' })
+  maxRow?: number;
```
muratcorlu, line 58: We didn't notice this detail in ADR phase but I think we should use plural form for this (`max-rows`) to be consistent with `rows` attribute. And it makes more sense to use plural here.

ozkersemih: I think your suggestion is meaningful so i changed as `max-rows`

muratcorlu:
@ozkersemih I just noticed that we don't have any storybook demo for disabled state of Textarea. It's reported that disabled state of the Input component is broken now (#387) To avoid missing this style, please check disabled style and provide one or a few stories to demonstrate disabled state in storybook.

fatihhayri (approved):
ok.

muratcorlu:
@buseselvi I added new stories for long labels in textarea. Also I changed many other stories to make visual effects of the parameters more visible in image snapshots.

Also noticed that we are not getting snapshots test for focus and invalid states of our form fields including input, textarea and select. We already have a separate issue for this (#207) and I'll rise the priority for it. So I think it's wise to fix that issue for textarea also on that issue.

muratcorlu (approved):

leventozen (approved):
Great work! I really liked the component you created and I appreciate the approach you took with the CSS. I would like to congratulate you on your work and mention that this component can be used as an example for future components.

Issue #362, opened 2023-01-09
npx commands in GitHub actions


muratcorlu:
We use `npx` commands in some of our CI/CD jobs to run some apps like semantic-release and playwright. `npx` is always using latest version of those libraries if we don't specifically mention a version. 2 days ago semantic-release, released [their version v20](https://github.com/semantic-release/semantic-release/releases/tag/v20.0.0) by dropping support from node v16 and this [broke our pipeline](https://github.com/Trendyol/baklava/actions/runs/3872612055/jobs/6601716142).

We need to have a decision about how to mitigate that kind of issues.

Issue #361, opened 2023-01-09
Node 18 support


muratcorlu:
Currently some parts of the Baklava development environment doesn't work with node 18. Since [Node 18 is the new LTS version](https://nodejs.org/en/blog/release/v18.12.0/), this needs to be fixed.

muratcorlu:
Blocker part of this issue is Storybook. Current version of Storybook doesn't work properly with node 18 (https://github.com/storybookjs/storybook/issues/21739). And apparently this will be fixed in storybook v7, which is in pre-release now. There is a workaround as using legacy provider but since node 18 version support is not that much critical for now and storybook v7 stable release seems to be released very soon, I'm eager to waiting for it.

Also `source-map` dependency needs to be upgraded. (https://github.com/mozilla/source-map/issues/454)

PR #363, opened 2023-01-09
ci: semantic-release node 18 requirement fix

muratcorlu:
2 days ago semantic-release, released [their version v20](https://github.com/semantic-release/semantic-release/releases/tag/v20.0.0) by dropping support from node v16 and this [broke our pipeline](https://github.com/Trendyol/baklava/actions/runs/3872612055/jobs/6601716142).

This is quickly fixing this issue by using v19 of semantic-release in our pipeline.

I created to issues about [Node 18 Support](#361) and [npx commands in our pipelines](#362) to track this issue and create a permanent solution, but for now I'm proposing this quick solution since this problem currently blocks our releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,7 +20,7 @@ jobs:
           node-version: 16
       - name: Install dependencies
         run: npm ci
-      - name: Create env file with release config 
+      - name: Create env file with release config
         run: |
           touch .env
           echo RELEASE=true >> .env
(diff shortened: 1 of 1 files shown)


leventozen (approved):
Since we don't dependent on the semantic-release in our code base, only to release, we can move forward with this solution for now. We will have a permanent solution after fix the issues that you mentioned. Necessary fix, good job!

PR #365, opened 2023-01-11
docs: add requirements documentation

### LINKED ISSUE #304: Documentation - Requirements & Polyfills
As we discussed here https://github.com/Trendyol/baklava/discussions/232, there are some limitations and requirement to use baklava in browsers, more specifically for `Lit.` And, right now there is no proper way to insert polyfills lazly inside Baklava without changing the bundle size. Also, user take this responsibility to decide which browsers or browser versions should my project supports.
Therefore, we should address this in our documentation. There is a very good example in [Lit.dev](https://lit.dev/docs/tools/requirements/).

leventozen:
In this documentation, we want to guide baklava users how they improve their browser support and to specify baklava's browser support by default. Right now, we support **95%** of all browser versions by default.

Why we avoid by adding polyfills to the baklava? Because, there is no way to dynamically add polyfills in baklava that runs only when needed. We don't want to have a fat bundle.

During the work, we found out, there are some other options. Actually, these options that came our mind before, but needed so much effort that we didn't have at that time when we prioritized issues. 
We may add some additional configs and strategies to improve our default bundle, or create a bundle that supports legacy browsers. We will create regarding issues about that. 

Stay tuned!

Closes #304.

### DIFF
--- a/docs/requirements.stories.mdx
+++ b/docs/requirements.stories.mdx
@@ -14,22 +14,22 @@ import { Meta } from '@storybook/addon-docs';
 
 # Requirements
 
-Baklava is a [web component](https://developer.mozilla.org/en-US/docs/Web/Web_Components) based design system. We are using [Lit](https://lit.dev) to simplify the process of creating web components and we create an [ECMAScript module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) (esm).
+Baklava is a [web component](https://developer.mozilla.org/en-US/docs/Web/Web_Components) based design system. We are using [Lit](https://lit.dev) to simplify the process of creating web components and we create an [ECMAScript module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) (ESM).
 It is built on top of web standards and it should work without any issues in all modern browsers, but it's important to note that for older versions of the browsers,
 the support for web components may be limited or non-existent, this is where polyfills can be used to fill the gaps.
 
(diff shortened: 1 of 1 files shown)


docs/requirements.stories.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+Some common examples of polyfills for web components are:
+- Custom Elements API: As mentioned before, older browsers do not have native support for the Custom Elements API, which is required to define custom elements. The `custom-elements-es5-adapter.js`` polyfill can be used to support custom elements in these older browsers.
+- Shadow DOM: As previously explained, the Shadow DOM allows web components to have their own encapsulated styles and layout. However, older browsers do not have native support for the Shadow DOM. The `shadydom.min.js` and `shadycss.min.js` polyfills can be used to support the Shadow DOM in older browsers.
+- HTML Templates: The template element is used to define the template for a web component. However, older browsers do not have native support for this element. The `template.js` polyfill can be used to support the template element in older browsers.
+- Babel Polyfill: This is a widely used polyfill that implements a large subset of the ECMAScript 6 standard, including support for modules. You can import it on your code with import `@babel/polyfill`.
```
muratcorlu, line 73: [Their documentation says](https://babeljs.io/docs/en/babel-polyfill) this is deprecated and they suggest to use [core-js](https://github.com/zloirock/core-js).

muratcorlu (approved):
Good job! 🚀

Issue #173, opened 2022-07-28
PR pipeline fails with PRs coming from forks


muratcorlu:
Our PR pipeline starts with `pull_request` event for the PRs those targets `next` branch. When we get a PR from a fork, this pipeline doesn't set secrets. Then this causes failure for Sonar job. We noticed that issue first time with #154. I changed `pull_request` event to [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) and it worked fine. But 20 days later I realized that, `pull_request_target` uses `next` branch as the source. That means during that time, non of the PR tests actually run over related branch. Then recently [I reverted back](https://github.com/Trendyol/baklava/commit/0d986ef4ea6021fc886cf2ff4213b8b5bb867694) this change.

At this point, we again in the position that we can not get PRs from forks since sonar job fails immediately. We need to find a solution for this.

PR #371, opened 2023-01-16
feat(input): step feature added for number input

### LINKED ISSUE #370: Input Component Step Feature
For input components that has `type='number'` attribute; Increase and decrease feature runs one by one with keyboard and input arrow keys. I want to Increase and decrease by specific step number instead of one by one. 
At the below gifs keyboard arrow keys used to increase and decrease. 
Baklava Playground (1)

May be set additional step prop to input component and provide native step experience.
Baklava Playground

**Describe alternatives you've considered**
My best advice; use native step feature.

**Additional context**
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step

yasin749:
Closes #370


yasin749:
#370

yasin749:
#370

yasin749:
#370

src/components/input/bl-input.stories.mdx
```diff
@@ -85,6 +89,7 @@ export const SingleInputTemplate = (args) => html`<bl-input
     maxlength='${ifDefined(args.maxlength)}'
     min='${ifDefined(args.min)}'
     max='${ifDefined(args.max)}'
+    step='${ifDefined(args.max)}'
```
muratcorlu, line 92:
```suggestion
    step='${ifDefined(args.step)}'
```

muratcorlu (approved):
Hi @yasin749 Thanks for the contribution. That's a very good addition.

There is a problem in our pipeline that we hit multiple times (mentioned in #173). This time I want to fix that issue instead of walking around it. I'll create a PR to fix that issue first, then we can merge your work.

yasin749:
Thank you @muratcorlu

PR #377, opened 2023-01-17
ci: pr worklow rearranged

### LINKED ISSUE #173: PR pipeline fails with PRs coming from forks

muratcorlu:
This issue was blocking us many times and we needed to work around it. Fixing this by keeping Sonar Cloud could be done [with a workaround](https://community.sonarsource.com/t/sonar-cannot-be-run-on-pr-from-a-fork/69229) but testing this is not easy. To unblock new contributions, I removed SonarCloud job from the workflow and enabled GitHub CodeQL as an alternative. Sonar Cloud can be enabled in the future once we have a safe and clean solution.

I also needed to do some changes to how we deploy our storybook/chromatic.

Fixes #173


olkeoguz (approved):
@muratcorlu the workaround you shared also seems to work just fine. May I ask what was our concern about that and why is it hard to test ? It says once the first job (build) finishes, the second job (Sonar step) downloads and scans the covarege from it.

Also does SonarCloud have an issue related to our problem so that we can follow it from there ? Or isn't it possible to go forward with CodeQL till the end.

muratcorlu:
@olkeoguz I don't know an easy way to test CI/CD changes for pull requests since they don't work in PRs. They needs to be pushed to main branch to start using them (for safety reasons). To be able to test that workaround, I need to have 2 github users, a separate repository and an account for Sonar for that repo. And then we will hope same solution will just work for baklava. Maybe we can do all of these anyway but for a beginning step to unblock current PR, I thought it's ok to just remove Sonar job for now.

I already enabled CodeQL and maybe indeed we'll even not need to re-enable sonarqube if CodeQL will be enough for our needs.

PR #381, opened 2023-01-19
feat(input): implement password type to input

### LINKED ISSUE #302: Input's `name` property not present on TypeScript
**Describe the bug**
`<BlInput />` doesn't support "name" property however it was [referenced in the document](https://trendyol.github.io/baklava/?path=/docs/components-input--text-input) for usage in forms

**Expected behavior**
Should accept "name" property

**Versions**
Baklava v2.0.0-beta.39
React 18.2.0

AykutSarac:
Added `password` type for bl-input, added toggle action for text visibility.
Additionally, closes #302


muratcorlu:
Thanks for the contribution @AykutSarac ! Please update tests to cover new functionality.

AykutSarac:
@muratcorlu Done! Let me know if anything is missing.

src/components/input/bl-input.ts
```diff
@@ -196,29 +210,45 @@ export default class BlInput extends FormControlMixin(LitElement) {
...
       ? html` <bl-icon class="custom-icon" name="${this.icon}"></bl-icon>`
       : '';
     const label = this.label ? html`<label for="input">${this.label}</label>` : '';
 
+    const passwordShowHide = this.passwordInput
+      ? html` <bl-icon
+          class="password-icon"
+          @click="${this.textVisiblityToggle}"
+          name="${this.passwordVisible ? 'eye_off' : 'eye_on'}"
+        ></bl-icon>`
```
muratcorlu, line 236: 1. Instead of making icon component clickable, we use icon-only button component (small, tertiary here) to make it accessible and consistent.
2. Since SVG icons are loaded lazily, to prevent showing an empty icon while it's loading when user toggles the button, put two separate buttons beforehand and show/hide with css. We already had similar issue on Select component before #256

src/components/input/bl-input.css
```diff
@@ -138,6 +138,11 @@ label {
   color: var(--bl-color-danger);
 }
 
+.password-icon {
+  cursor: pointer;
+  user-select: none;
+}
```
muratcorlu, line 144: If you use regular button component, you'll not need those styles.

muratcorlu:
Apparently, Microsoft Edge has its own password reveal button by default. We may consider hiding it for the consistency https://learn.microsoft.com/en-us/microsoft-edge/web-platform/password-reveal

muratcorlu (changes requested):
I also realized that, current implementation can have issues with the buttons placed by password managers (like icloud keychain or 1password) To avoid this, I think we need to put icons next to the input instead of putting it on top of input with absolute positioning. But this will need a bigger refactor on the component because currently we use internal input for the borders of the component. Probably we'll need to wrap internal input with another element and handle positioning icons in it.

src/components/input/bl-input.css
```diff
@@ -115,51 +165,42 @@ label {
   padding: 0;
 }
 
-.dirty:invalid ~ label {
-  color: var(--bl-color-danger);
+::part(button) {
+  --padding-horizontal: 5px;
+  --text-hover-color: transparent;
 }
```
muratcorlu, line 171: Opening a `part` means this internal button element can be styles in any way outside of the component. I don't think that we want this. I see that you wanted to override hover background and paddings. But we should either not need this or should fix in another way.

I think a standart `tertiary`, `small` and `neutral` icon only button should be fine for here.

AykutSarac: Is there any way to override button's style? If I keep the button size to medium I have to remove the hover background, otherwise I have to set the size to small but then the icon size will be smaller compared to the next icon to it.

src/components/input/bl-input.ts
```diff
@@ -239,7 +286,7 @@ export default class BlInput extends FormControlMixin(LitElement) {
           aria-describedby=${ifDefined(this.helpText ? "helpText" : undefined)}
           aria-errormessage=${ifDefined(this.checkValidity() ? undefined : "errorMessage")}
         />
-        <div class="icon">${icon}<bl-icon class="error-icon" name="alert"></bl-icon></div>
+        <div class="icon">${revealButton || icon}<bl-icon class="error-icon" name="alert"></bl-icon></div>
```
muratcorlu, line 289: With this implementation, it's not possible to see a custom icon next to password reveal button. So

```html
<bl-input type="password" icon="key"></bl-input>
```

will not work as expected. I suggest to have another wrapper for reveal button similar to `<div class="icon">` and don't touch the logic about custom icon here.

src/components/input/bl-input.ts
```diff
@@ -210,20 +228,49 @@ export default class BlInput extends FormControlMixin(LitElement) {
...
+          size="small"
+          kind="neutral"
+          variant="tertiary"
+          class="${classMap({
+            'reveal-button': true,
+            hide: !this.passwordVisible
+          })}"          aria-labelledby="reveal"
+          icon="eye_off"
+          @click="${this.textVisiblityToggle}"
+        ></bl-button>`
```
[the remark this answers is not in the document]
muratcorlu: I opened an issue for the localization part: #398

AykutSarac:
Updated accordingly, fyi @muratcorlu.

muratcorlu (approved):
We have a good neat work here. Congrats @AykutSarac

buseselvi:
Great work! We can use **neutral kind** icon button to open and close the eye like in the design (2nd image) in every input state except the default state, the button color in the default version looks ok 👌🏻 

This is because when we use the danger button for the eye in case of error it looks dangerous to press but it is not. @AykutSarac @muratcorlu 

1 image
2 image

muratcorlu:
Oh, good catch @buseselvi is it ok to use natural color in every case? Including focus state.

muratcorlu (approved):

PR #384, opened 2023-01-23
docs: adr and figma links have been added to component docs

muratcorlu:
* Added ADR and Figma links to component storybook pages.
* Added some useful links to our Readme file to make them easy to discover.
* I grouped sub components in a single folder to simplify the component list in storybook (like tab, tab-panel, tab-group to tab)

image


DamlaDemir:
`Figma` and `ADR` links are good ideas  👏    It would be nicer if they could be opened in a new tab but when I glanced at google I saw `git.md` doesn't support this feature.

DamlaDemir (approved):

PR #385, opened 2023-01-24
fix(select): reflect label-fixed attribute

AykutSarac:
Fixes, `labelFixed` property is unusable as a React attribute.


muratcorlu (approved):
Good catch! 💯

PR #395, opened 2023-02-01
fix(button): designs for focused state

### LINKED ISSUE #340: Focus state for Buttons
Our buttons currently don't have focus state. This causes problems when we want to make our elements with keyboard navigation.

Baklava Buttons should be able to focus by keyboard and should have a visual indication that is focused now.

pratikgaloria:
Fixes #340

### DIFF
--- a/src/components/button/bl-button.css
+++ b/src/components/button/bl-button.css
@@ -68,6 +68,19 @@
   outline: none;
 }
 
+.button:focus-visible {
+  position: relative;
+  outline: none;
+}
+
+.button:focus-visible::after {
+  border: 2px solid var(--main-color);
+  border-radius: var(--bl-border-radius-l);
+  content: "";
+  position: absolute;
+  inset: -4px;
+}
+
 :host ::slotted(bl-icon) {
   font-size: var(--icon-size);
 }


src/components/button/bl-button.css
```diff
@@ -68,6 +68,19 @@
   outline: none;
 }
 
+.button:focus-visible {
+  position: relative;
+  outline: none;
```
muratcorlu, line 73: When I see this, I just asked myself "why we are not just using outline and adding so many stuff like this" and realized that [there is a bug](https://bugs.webkit.org/show_bug.cgi?id=246192) in webkit about using `outline` with `border-radius`. It's actually fixed and release in [Safari Techonology Preview 157](https://webkit.org/blog/13575/release-notes-for-safari-technology-preview-157/) last year, but it's somehow not released yet in stable release. It's sad that this bug adds extra complication here.

muratcorlu (approved):

muratcorlu:
Apparently, "close button" of our dialogs get focus automatically when dialog is opened. And now a UI reviews is needed. We were not aware of this until now since there was no visual indicator. Dialog moves focus from document to dialog content and since close button is the first focusable element inside dialog, it gets focused. I'm not sure if we should just accept this or we need to consider moving focus to another element in dialog. Any ideas? @buseselvi @DamlaDemir 

image

muratcorlu:
@pratikgaloria It would be nice to have a story in our buttons storybook to demonstrate and have snapshots testing for the focus state. You may check `radio group` component's "Focused option" story as an example to have a snapshot test for focus state.

buseselvi:
> Apparently, "close button" of our dialogs get focus automatically when dialog is opened. And now a UI reviews is needed. We were not aware of this until now since there was no visual indicator. Dialog moves focus from document to dialog content and since close button is the first focusable element inside dialog, it gets focused. I'm not sure if we should just accept this or we need to consider moving focus to another element in dialog. Any ideas? @buseselvi @DamlaDemir
> 
> image

When i benchmark at other design systems,
- In Carbon and Atlassian, primary action is getting focused, then close button, then secondary button.
image

- In Nord, primary action is getting focused, too. Then with shift+tab secondary button and then close button is getting focused.
image

- In Gitlab, first button of the action area is getting focused first, the last close button becomes focused.
image

After looking at these examples, my suggestion is to open dialog with the primary action focused. But I wasn't sure which one should come after: close button or secondary button?

muratcorlu:
@buseselvi I checked this issue and looked for a quick solution in dialog component but apparently it'll need some further investigation and even maybe some technical changes on how we implement button component. I created a separate issue (#406). If you think that is not a big issue visually (there are some pages uses dialog on production, I think) I would suggest to release this PR without waiting fixing dialog focus.

buseselvi:
@muratcorlu we can release this one and fix the dialog later 👌🏻

PR #410, opened 2023-02-09
chore: track dependency updates

muratcorlu:
Sometime we are missing important updates on our dependencies. Dependabot can help about this.

Documentation about dependabot.yml file can be found here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file


olkeoguz (approved):

leventozen (approved):
We needed, good work! We may add `package-ecosystem: "github-actions"` in a same way to track its dependencies as well.

muratcorlu:
@leventozen good point, I'm adding it as well.

PR #412, opened 2023-02-10
style(input): style autofill input

### LINKED ISSUE #403: Autofill Styling
**Describe the bug**
When input is autofilled the styling looks unstatisfying.

**Expected behavior**
217243896-b2ad8c77-0fb7-41cb-84fe-e3729047fe37

**Screenshots**
image

**Additional context**
Baklava.2.0.0-beta.67

AykutSarac:
Fixes #403


src/components/input/bl-input.css
```diff
@@ -84,6 +89,16 @@ input {
   background-color: transparent;
 }
 
+input:disabled {
+  cursor: not-allowed;
+}
+
+input:-webkit-autofill,
```
muratcorlu, line 96: I think we need to use `:autofill` as well. It's supported by Firefox and Safari. Most probably vendor prefix will drop in the future.

src/components/input/bl-input.css
```diff
@@ -84,6 +89,16 @@ input {
...
+input:-webkit-autofill:hover,
+input:-webkit-autofill:focus {
```
muratcorlu, line 98: Do we really need to set hover and focus states as well?

src/components/input/bl-input.css
```diff
@@ -84,6 +89,16 @@ input {
...
+  box-shadow: 0 0 0 40rem var(--autofill-bg-color) inset;
```
muratcorlu, line 99: I see that it's not possible to just set background-color because of some browsers doesn't allow it. Since this is weird to see for future developers, I have 2 suggestions:

1. Use `background-color` anyway. It'll work for the browsers that allows setting it.
2. As an addition to using `background-color` add this box-shadow trick with a comment on top of it.

```suggestion
  background-color: var(--autofill-bg-color);
  
  /**
   * Some browsers doesn't allow setting background-color
   * https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill
   */
  box-shadow: 0 0 0 40rem var(--autofill-bg-color) inset;
```

src/components/input/bl-input.css
```diff
@@ -44,6 +45,10 @@
   --icon-size: var(--bl-font-body-text-3-line-height);
 }
 
+.wrapper:has(input:-webkit-autofill, input:-webkit-autofill:hover, input:-webkit-autofill:focus) {
+  box-shadow: 0 0 0 40rem var(--autofill-bg-color) inset;
```
muratcorlu, line 49: I understand the issue about overriding autofill style but since this is the wrapping element, you can just use `background-color`.

src/components/input/bl-input.css
```diff
@@ -44,6 +45,10 @@
   --icon-size: var(--bl-font-body-text-3-line-height);
 }
 
+.wrapper:has(input:-webkit-autofill, input:-webkit-autofill:hover, input:-webkit-autofill:focus) {
```
muratcorlu, line 48: Because of [browser support for `:has` selector](https://caniuse.com/?search=%3Ahas) not perfect yet, I see this will not make the experience of some users about auto filling very good. But I also see that, trying to cover older browsers will need some work on JS and this will add performance costs for every user. So I think it's ok to keep this in that way for now.

My comments below about vendor prefix and hover/focus selectors also applies here.

src/components/input/bl-input.css
AykutSarac: Not, apparently.

src/components/input/bl-input.css
```diff
@@ -44,6 +45,14 @@
   --icon-size: var(--bl-font-body-text-3-line-height);
 }
 
+.wrapper:has(input:autofill) {
+  background-color: var(--autofill-bg-color);
+}
+
+.wrapper:has(input:-webkit-autofill) {
+  background-color: var(--autofill-bg-color);
+}
```
muratcorlu, line 54:
```suggestion
.wrapper:has(input:autofill),
.wrapper:has(input:-webkit-autofill) {
  background-color: var(--autofill-bg-color);
}
```

src/components/input/bl-input.css
```diff
@@ -84,6 +93,30 @@ input {
...
+
+input:-webkit-autofill {
+  background-color: var(--autofill-bg-color);
+
+  /**
+   * Some browsers doesn't allow setting background-color
+   * https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill
+   */
+  box-shadow: 0 0 0 40rem var(--autofill-bg-color) inset;
+}
```
muratcorlu, line 118:
```suggestion
input:autofill,
input:-webkit-autofill {
  background-color: var(--autofill-bg-color);

  /**
   * Some browsers doesn't allow setting background-color
   * https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill
   */
  box-shadow: 0 0 0 40rem var(--autofill-bg-color) inset;
}
```

src/components/input/bl-input.css
AykutSarac: It doesn't work that way, so I seperated them.

src/components/input/bl-input.css
AykutSarac: Same as the comment above.

src/components/input/bl-input.css
muratcorlu: How could that be possible? 🤔 Weird.. I'll check the reason.

olkeoguz: The :has pseudo-class is not supported by Firefox and by recent (not so old) versions by chrome and safari. We may think of another ways to implement these.

https://caniuse.com/css-has

AykutSarac: You're right, the browser support is not enough yet to implement this. I'm open to suggestions for alternative solutions.

muratcorlu: I added my comments on browser support earlier:
> Because of [browser support for :has selector](https://caniuse.com/?search=%3Ahas) not perfect yet, I see this will not make the experience of some users about auto filling very good. But I also see that, trying to cover older browsers will need some work on JS and this will add performance costs for every user. So I think it's ok to keep this in that way for now.
> 
> https://github.com/Trendyol/baklava/pull/412#discussion_r1102737194

I have some ideas to make this without `:has` selector but I don't see any way that doesn't effect the users who has a modern browser (85% of all global users) in terms of performance.

muratcorlu (approved):
I think we can go with this for now. If we can find a better way to "not repeat" styles, we can improve it.

PR #422, opened 2023-02-13
feat(input): add loading state

AykutSarac:
- Added button loading state
- Connected to disabled state, enables when loading is `true`
- Tests will be added/fixed after PR review

Notes:
- Hesitant about the relation between disabled and loading states handling
- Does loading state requires `reflect` property?

https://user-images.githubusercontent.com/47941171/218412709-69e30806-8b15-4c1f-ace2-b7da229f91d8.mov

### DIFF
--- a/src/components/button/bl-button.css
+++ b/src/components/button/bl-button.css
@@ -1,3 +1,8 @@
+@keyframes spin {
+  from { transform: rotate(0deg); }
+  to { transform: rotate(359deg); }
+}
+
 :host {
   display: var(--bl-button-display, inline-block);
   max-width: 100%;
@@ -23,7 +28,7 @@
   justify-content: var(--bl-button-justify, center);
   align-items: center;
   box-sizing: border-box;
-  width:100%;
+  width: 100%;
   height: var(--height);
   border: solid 1px var(--border-color);
   border-radius: 6px;
@@ -76,9 +81,13 @@
 .button:focus-visible::after {
   border: 2px solid var(--main-color);
   border-radius: var(--bl-border-radius-l);
-  content: "";
+  content: '';
   position: absolute;
   inset: -4px;
+}
+
+.loading-icon {
+  animation: spin 1s linear infinite;
 }
 
 :host ::slotted(bl-icon) {
--- a/src/components/button/bl-button.ts
+++ b/src/components/button/bl-button.ts
@@ -50,6 +50,12 @@ export default class BlButton extends LitElement {
   @property({ type: String })
   label: string;
 
+   /**
+   * Sets loading state of button
+   */
+   @property({ type: Boolean, reflect: true })
+   loading = false;
+
   /**
    * Sets button as disabled
    */
@@ -151,8 +157,10 @@ export default class BlButton extends LitElement {
   }
 
   render(): TemplateResult {
+    this.disabled = this.loading;
     const isAnchor = !!this.href;
     const icon = this.icon ? html`<bl-icon name=${this.icon}></bl-icon>` : '';
+    const loadingIcon = this.loading ? html`<bl-icon class="loading-icon" name="loading"></bl-icon>` : '';
     const slots = html`<slot name="icon">${icon}</slot> <span class="label"><slot></slot></span>`;
     const caret = this.dropdown ? this.caretTemplate() : '';
     const classes = classMap({
@@ -179,7 +187,7 @@ export default class BlButton extends LitElement {
           ?disabled=${this.disabled}
           @click="${this._handleClick}"
         >
-          ${slots} ${caret}
+         ${loadingIcon} ${slots} ${caret}
         </button>`;
   }
 }
--- a/src/components/button/bl-button.ts
+++ b/src/components/button/bl-button.ts
@@ -53,7 +53,7 @@ export default class BlButton extends LitElement {
    /**
    * Sets loading state of button
    */
-   @property({ type: Boolean, reflect: true })
+   @property({ type: Boolean })
    loading = false;
 
   /**
--- a/src/components/button/bl-button.test.ts
+++ b/src/components/button/bl-button.test.ts
@@ -86,6 +86,23 @@ describe('bl-button', () => {
 
       expect(el.getAttribute('target')).to.eq('_self');
     });
+
+    it('is disabled button during loading state', async () => {
+      const el = await fixture<typeOfBlButton>(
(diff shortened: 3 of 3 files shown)


muratcorlu:
@AykutSarac Is there a blocker to have progress on this?

AykutSarac:
> @AykutSarac Is there a blocker to have progress on this?

I couldn't identify the issues with the tests

muratcorlu:
I think this doesn't work as expected in `pagination` tests:

```ts
this.toggleAttribute('disabled', this.loading || this._disabled);
```

Instead of having another state for disabled, I would set disabled property directly once `loading` changes to `true`. I think this will be a safer approach.

src/components/button/bl-button.ts
```diff
@@ -50,6 +50,18 @@ export default class BlButton extends LitElement {
...
+   * Sets the button label for loading status.
+   */
+  @property({ type: String, attribute: 'loading-label' })
+  loadingLabel: string;
+
+   /**
+   * Sets loading state of button
+   */
+   @property({ type: Boolean })
+   loading = false;
```
muratcorlu, line 63: add `reflect: true` to parameters, since this is used in style queries.

And I think there is a problem with indentation here.

AykutSarac:
> Instead of having another state for disabled, I would set disabled property directly once loading changes to true. I think this will be a safer approach.

We should protect the initial state of disabled, that way we can know if the button is actually disabled or disabled due to loading state. This will avoid button from removing disabled state in case button is really disabled however has loading state too.

muratcorlu:
Good point. But do we really need to set disabled property to achieve this? Maybe we can consider using loading state to style and behave the component as in disabled state too. Like:

```
?disabled=${this.disabled || this.loading}
```

And

```css
:host([disabled]), :host([loading]) {
   /* disabled style */
}
```

pratikgaloria:
@AykutSarac @muratcorlu 

Shall we really allow changing the width of the actual component after the loading state is applied? When this component is wrapped by a parent component that restricts the flex, there will be an overflow or space-squeezing with sibling elements and I think it is not desired experience. wdyt?

muratcorlu:
> Shall we really allow changing the width of the actual component after the loading state is applied? When this component is wrapped by a parent component that restricts the flex, there will be an overflow or space-squeezing with sibling elements and I think it is not desired experience. wdyt?

Yes, that was also my hesitation point and we discussed about this in one of our previous design meetings. In that meeting we concluded that this is a very common approach and setting loading state is already under control of the app developer. So they will be aware of this re-flow. We also realized that, even without an icon, it's wise and recommended to change the label in loading state like form "Save" to "Saving..." and this showed us that a re-flow is inevitable.

In addition to those, there are still some ways of avoiding this like setting a bigger width, using a block button or using an icon-only button etc.

AykutSarac:
> Good point. But do we really need to set disabled property to achieve this? Maybe we can consider using loading state to style and behave the component as in disabled state too. Like:
> 
> ```
> ?disabled=${this.disabled || this.loading}
> ```
> 
> And
> 
> ```css
> :host([disabled]), :host([loading]) {
>    /* disabled style */
> }
> ```

I couldn't achieve this result, maybe I'm missing something.

src/components/button/bl-button.css
```diff
@@ -76,15 +81,24 @@
...
+  font-size: var(--icon-size);
+}
+
 :host ::slotted(bl-icon) {
   font-size: var(--icon-size);
 }
 
+:host([loading]) ::slotted(bl-icon) {
+  display: none;
+}
```
muratcorlu, line 100: Apparently this doesn't work with `icon` attribute, only works by setting icon slot.

image

Maybe we can cover icon slot with an element to target it regardless of where it set from.

src/components/button/bl-button.stories.mdx
```diff
@@ -189,6 +205,22 @@ If button has a limited width and a long text that can not fit in a single line,
...
+  <Story name="Loading Button With Icon" args={{ loading: true, loadingLabel: 'Logging in...', content: 'Login', icon: 'account' }}>
+    {LoadingStateTemplate.bind({})}
+  </Story>
+  <Story name="Loading State Buttons" args={{ loading: true, content: 'Login' }}>
+    {SingleButtonTemplate.bind({})}
+  </Story>
+  <Story name="Loading Button Custom Label" args={{ loading: true, loadingLabel: 'Logging in...', content: 'Login' }}>
+    {SingleButtonTemplate.bind({})}
+  </Story>
+</Canvas>
```
muratcorlu, line 222: Can you add a few more examples:

* An icon only button, without label.
* `small` and `large` examples.
* `disabled` example. You may consider to merge `disabled` with one of above examples, since we don't expect any visual change on `disabled` state.

These are for preventing us to break any of those states accidentally.

src/components/button/bl-button.stories.mdx
```diff
@@ -77,6 +85,13 @@ ${SingleButtonTemplate({size: 'large', ...args})}
 ${SingleButtonTemplate({size: 'medium', ...args})}
 ${SingleButtonTemplate({size: 'small', ...args})}`
 
+export const LoadingStateTemplate = (args) => html`
+<bl-button icon="play" @click="${(event)=>event.currentTarget.toggleAttribute('loading')}" icon="${ifDefined(args.icon)}"></bl-button>
```
muratcorlu, line 89: If we want to keep this interaction in story, then we need to have a play function to be able to take screenshot of this button in loading state in chromatic. Currently it looks like below:

image

Another option would be just putting `SingleButtonTemplate` like other examples here.

src/components/button/bl-button.stories.mdx
```diff
@@ -189,6 +204,16 @@ If button has a limited width and a long text that can not fit in a single line,
   </Story>
 </Canvas>
 
+## Loading Buttons
+
+Custom loading text can be set with `loading-label` attribute.
```
muratcorlu, line 209:
```suggestion
Button can be set in loading state. In this state button becomes disabled with a loading indicator. You can set this state by setting `loading` attribute.

A custom loading text can be also set with `loading-label` attribute. It's suggested to use `loading-label` to inform the user about the process.
```

muratcorlu:
Disabled tertiary buttons are broken with this change. Please check it.

image

muratcorlu (approved):
Good job!

buseselvi (approved):

PR #427, opened 2023-02-15
feat(icon): add star_double icon

umitkucuk:


muratcorlu:
I would name this icon `stars`. 😊

buseselvi:
Hi @umitkucuk our two star icon in the design file is below (i added the svg export for you). Can you replace it with the name "two_stars"?
Two_Star

buseselvi:
@muratcorlu we also have three stars icon with the name "rating", maybe we can name it as three_stars? 😅

umitkucuk:
I thought prefixing the icon names with "star_" would be good for the alphabetic order;

`star`, `star_double`, `star_triple` (If there is no icon with more than three stars 😊)

What do you think? @muratcorlu @buseselvi

umitkucuk:
It's ready for review again.

buseselvi (approved):

muratcorlu:
We didn't rename `rating` icon here but I'm also a bit hesitant about renaming it since there can be some people who already use it. But to be honest, I see more icons those have names with context in trendyol, which does not make sense for a generic library. I'll create a separate issue for that.

muratcorlu (approved):

muratcorlu:
@buseselvi I couldn't see this new icon in our [Figma document](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=140%3A3340). Can you sync it with same name?

Issue #323, opened 2022-11-21
Popover Component


leventozen:
TBD

buseselvi:
[Design Documentation](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=3369%3A26076&t=9atyD6p1ZMzRgKUl-1)

muratcorlu:
Good job @ozkersemih! Here are my points:

* `target` property can accept a direct reference of the target dom element as well as just id of it. So in frameworks we can take advantage of setting `target` property with a reference of an element without needing to give a uniquie id to it.
* As a principle, it would be great to accept all of the "visual" parameters as CSS variables. For me `arrow`, `offset`, `inline`, `shift` and `fitSize` are all visual parameters (doesn't change functionality). I know sometimes it's technically very difficult to take them as CSS variables since they are used in JS code as parameters of Floating UI, but here it would be great to be sure that: they are really needed and they need to be set as attributes instead of CSS variables. I can also check for the opportunities to use them as CSS variables but for arrow, I'm pretty sure we can take a `--bl-popover-arrow` as the `display` value of arrow element inside. So if developer doesn't want to have arrow, they can set it as `--bl-popover-arrow: none;`.
* I think should have some public methods like `show()` and `hide()`
* Popover can be triggered by a click on target element and with a `mouseenter` as well (for tooltip). I think we need extra arguments to define those events (with some nice defaults).
* In some use cases we want to sync width of popover with it's trigger element (select, dropdown). I think we'll need a parameter for enabling this.

ozkersemih:
Hi @muratcorlu thanks for your review.
* We can accept target as HTML element, its okay.
* We will try accept visual parameters with CSS variables as possible as.
* I think we don't need public `show` and `hide` method. We can already have `open` attribute and we can trigger `show` and `hide` methods with this attribute.
* We think popover is utility component. So maybe it doesn't need any attribute about `mouseenter` or `click`. 
  * e.g we can handle `mouseenter` and `mouseleave` in `tooltip` component. These methods can trigger popovers open attribute.
* We defined parameter `fitSize` for that.

muratcorlu:
> * We think popover is utility component. So maybe it doesn't need any attribute about `mouseenter` or `click`.
>   
>   * e.g we can handle `mouseenter` and `mouseleave` in `tooltip` component. These methods can trigger popovers open attribute.

[...]

PR #431, opened 2023-02-17
feat(select): native form integration and a11y improvements

### LINKED ISSUE #287: Using Select component in a form
Select component needs to have form field capabilities like `name`, `value` attributes and ElementInternals implementation.

### LINKED ISSUE #288: Accessibility improvements on Select component
- [ ] Select component should use related `aria-` attributes
- [ ] Select input should be usable with the keyboard. We need to follow a11y standards for the navigation approach.

### LINKED ISSUE #317: The select component does not update when one of the select options is selected programmatically
**Describe the bug**
When I change the selected options by state, the select component itself does not update.

**To Reproduce**
1. Decide which option should be selected with a state
2. Change the state from somewhere else
3. The select component does not get an update

**Expected behavior**
When I select the options programmatically (via the state), the select component itself should get an update according to the state changes.

**Screenshots**
https://codesandbox.io/s/react-typescript-forked-0fh8qn?file=/src/App.tsx

### LINKED ISSUE #341: Select doesn't work properly inside Drawer
**Describe the bug**
If you put a `bl-select` inside `bl-drawer`, popover of select becomes invisible.

**To Reproduce**

I made an example here: https://codepen.io/muratcorlu/pen/OJEeKLr

**Expected behavior**
Popover of select should work as normal inside Drawer. It works properly within Dialog.

**Desktop:**
 - OS: macOS
 - Browser Safari
 - Version 16.2

### LINKED ISSUE #396: Double @bl-change event emitted in pagination component with multiple instances.
**Describe the bug**
When multiple instances of the pagination component are used, a single page change action triggers multiple @bl-change events, instead of the expected single event. This issue may affect the functionality and accuracy of applications using the affected component.

**To Reproduce**
1. Go to https://codesandbox.io/s/vue-2-playground-forked-buxp7u?file=/src/components/Pagination.vue
2. Click on pagination buttons
3. See pagination change event count increase twice every time
4. Then click remove second pagination checkbox and click pagination buttons again
5. See pagination change event count increase one by one.

**Expected behavior**
It should emit an event only once, regardless of the number of instances present.

 - OS: MacOS
 - Browser: chrome

### LINKED ISSUE #426: Alignment with error message
**Describe the bug**
I want to align the button and input next to each other, label-fixed enabled. However when the error message is shown the alignment is getting broken.

**Expected behavior**
Alignment should be kept nonetheless. 

**Screenshots**
https://user-images.githubusercontent.com/47941171/219015156-4637cadf-420e-4eb8-b055-3f8fecf9f94f.mov

 - Version 71

### LINKED ISSUE #428: Baklava input, select etc. element problem. Not working BS modal
image

Baklava Select component works wrong on BS modal.
We need full size input element. (width:100%) Sizes are restricted. Not working full width

muratcorlu:
* This PR adds ElementInternals support for Select component. So it can be used as a form field by setting/reading `name` and `value` properties.
* Select `value` can be a `string` and `string[]`. Also it's possible to use it with arbitrary types. I add examples to the storybook.
* Also added keyboard navigation. Now Select component gets focus by tabbing over it. `Enter` or `Space` toggles select popover. Once popover is open navigation between options is possible with `ArrowDown` and `ArrowUp` keys. `Escape` button closes popover.
* CSS `transform: translateX(0)` was breaking popover positining inside Drawer component. That was causing issues like #341 Fixed it by positioning drawer by playing with `right` attribute only. I also added a CSS variable to select named `--bl-popover-position` to be able to customize popover positioning with CSS. I named it as `--bl-popover` intentionally because it needs to be defined in [Popover component](#323). Once select implemented with new Popover, this will not be a property of Select anymore.
* I fixed pagination component to use new form of Select for its "items per page selection".
* Fixed spacing issue on help-text and error message (aligned with textarea spacing)

Closes #287 
Fixes #426 
Fixes #317 
Fixes #288 
Fixes #341 
Fixes #428 
Fixes #396


AykutSarac:
Maybe an improvement here can be made to allow value to be number type?

muratcorlu:
There is an issue about showing custom error message on form submit. I'm getting help about this issue here: https://github.com/open-wc/form-participation/issues/39

Other than that, this PR mostly works as expected. I'm working on improving test cases and storybook documentation.

muratcorlu:
> Maybe an improvement here can be made to allow value to be number type?

With this PR, you can use any type as the value (with JS). For example:

```html
<bl-select label="Products" label-fixed name="product" id="productSelect" multiple></bl-select>

<bl-button id="addProduct">Add Product Option</bl-button>

<script>
      document.getElementById('addProduct').addEventListener('click', () => {
        const id = Math.floor(Math.random() * 1000);
        const product = {
          id,
          name: `Product ${id}`,
          price: `$${id}`,
          toString() {
            return this.id;
          }
        };

        const option = document.createElement('bl-select-option');
        option.value = product;
        option.textContent = product.name;
        document.getElementById('productSelect').appendChild(option);
      });

      document.getElementById('productSelect').addEventListener('bl-select', (e) => {
        console.log(e.detail);
      });
</script>
```

Here in the example, we are setting complex objects (Product objects) as the value for options. And once you do selection, you get references to the selected objects. I think this will be nice for advanced use cases like in React.

image

And this is `event.detail` once you made a selection: 

image


@AykutSarac Does this gives enough confidence about the value type?

muratcorlu:
With TypeScript, you can do this:

```ts
const option = document.createElement<BlSelectOption<number>>('bl-select-option');
```

So `BlSelect` and `BlSelectOption` takes the type of the value as the generic type.

AykutSarac:
Any ETA for this to release?

muratcorlu:
> Any ETA for this to release?

Remaining items that I'm working on:

1. Issue about submitting the form needs to have a PR on related dependency: https://github.com/open-wc/form-participation/issues/39 But I don't think that this PR needs to be blocked with it. Because this issue is already there for other form fields too. Fixing it will fix all of them.
2. I need to improve unit tests
3. I need to improve storybook documentation with more stories and examples
4. And more importantly, I need to have *reviews* 😊 

I think we can still manage to merge this next week.

src/components/drawer/bl-drawer.css
```diff
@@ -1,19 +1,9 @@
...
-  }
-}
-
 .drawer {
   box-sizing: border-box;
   position: fixed;
   display: flex;
   top: 0;
-  right: 0;
+  right: -100%;
```
AykutSarac, line 6: Why did you switch from `transform: translate` to absolute positioning here?

scripts/generate-react-exports.js
```diff
@@ -49,20 +53,34 @@
...
+    const typeParam = componentDeclaration.typeParameters.map((param) => `${param.name}`).join(', ');
+
+    const compType = `${Type}<${typeParam}>`;
+    const wrapperCompType = `ReactWebComponent<${compType}${eventTypes}>`;
+
+    baklavaReactFileParts.push(`export function ${componentName}<${typeParamDefinition}>(props: React.ComponentPropsWithRef<${wrapperCompType}>): React.LazyExoticComponent<${wrapperCompType}> {
+      return React.lazy(() => ${componentDefinition(`<${compType}>`)} )(props);
+    }`);
+  } else {
+    baklavaReactFileParts.push(`export const ${componentName} = React.lazy<ReactWebComponent<${Type}${eventTypes}>>(() => ${componentDefinition(`<${Type}>`)} );`);
```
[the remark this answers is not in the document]
muratcorlu: I don't see any risks here, since the source of this JSON object is always the output of our custom-element-analyzer. Dismissing this.

pratikgaloria:
@muratcorlu @AykutSarac 

I have a doubt whether the object as a value will work here the way we want it to, because when you provide an object as a value to the option and to the `value` property of <bl-select>, that option won't be added to `selectedOptions` until the value respects reference equality, because we are calculating selectedOptions with `this.value === option.value` and it will be false always. Also, the same with Array.includes() and it won't work there as well.

e.g.
```jsx
<bl-select value={{a: 1}}>
  <bl-select-option value={{a: 1}}>${option.label}</bl-select-option>
</bl-select>
```
the above won't add `{a: 1}` into `selectedOptions` I believe. while for many use-cases I can think of, there will be a need to programmatically supply a value to the select to perform the initial options selection. wdyt?

muratcorlu:
@pratikgaloria Complex value support is useful if you work with references. When you have a list of `User` object and want to show a selection over them, instead of using string IDs you are able directly use object itself. In that case, when user do a selection, you will get a reference of those objects as the value.

Of course it will not work by giving completely independent (but similar objects) as the value of select and option. But I think there should be no expectation like that. Generally, we set values like:

```html
<bl-select value={users[0]}>
  <bl-select-option value={users[0]}>${option.label}</bl-select-option>
  <bl-select-option value={users[1]}>${option.label}</bl-select-option>
</bl-select>
```

And the target here is giving that "option". By default select value is `string | string[]`. If you want to use with a different type, then you'll be more aware of what you are doing, I think.

src/components/select/bl-select.ts
```diff
@@ -241,58 +330,74 @@ export default class BlSelect extends LitElement {
...
+  private focusedOptionIndex = -1;
+
+  private handleKeydown(event: KeyboardEvent) {
+    if (this.focusedOptionIndex === -1 && event.code === 'Enter' || event.code === 'Space') {
+      this.togglePopover();
+    } else if (event.code === 'Escape') {
+      this.close();
+      event.preventDefault();
+    } else if (this._isPopoverOpen && ['ArrowDown', 'ArrowUp'].includes(event.code)) {
+      event.code === 'ArrowDown' && this.focusedOptionIndex++;
```
pratikgaloria, line 356: I noticed one thing here that when you press arrow keys it also scrolls the page along with the option navigation. Is this expected? somehow it gives inconsistent effect to the users. maybe shall we `event.preventDefault()` and then only navigate the options?

pratikgaloria: Also, after pressing the `Space` it scrolls me down of the page, I think we should preventDefault there too.

src/components/select/bl-select.ts
```diff
@@ -241,58 +330,74 @@ export default class BlSelect extends LitElement {
...
     </div> `;
   }
 
-  private get _showPlaceHolder() {
-    if (this.label && !this.labelFixed) {
-      return !this._selectedOptions.length && this._isPopoverOpen;
+  private focusedOptionIndex = -1;
+
+  private handleKeydown(event: KeyboardEvent) {
+    if (this.focusedOptionIndex === -1 && event.code === 'Enter' || event.code === 'Space') {
```
pratikgaloria, line 350: How about having `ArrowDown` as well to toggle the popup specially for this component?

src/components/select/bl-select.ts
```diff
@@ -241,58 +330,74 @@ export default class BlSelect extends LitElement {
...
+      <div class="hint">${invalidMessage} ${helpMessage}</div>
     </div> `;
   }
 
-  private get _showPlaceHolder() {
-    if (this.label && !this.labelFixed) {
-      return !this._selectedOptions.length && this._isPopoverOpen;
+  private focusedOptionIndex = -1;
+
+  private handleKeydown(event: KeyboardEvent) {
```
pratikgaloria, line 349: looks like for the multiple select, keyboard selection is not working properly. I tried with arrow keys and then pressing Enter/Space on the options but it just toggles the popover rather than selecting the options, maybe checkbox is not getting proper focus?

src/components/select/bl-select.ts
```diff
@@ -241,58 +330,74 @@ export default class BlSelect extends LitElement {
...
   }
 
-  private _onClickSelectInput() {
+  private togglePopover() {
     this._isPopoverOpen ? this.close() : this.open();
   }
 
   private _handleSelectEvent() {
-    this._onBlSelect(this._selectedOptions);
+    this._onBlSelect(this._selectedOptions.map(option => ({
```
pratikgaloria, line 375: Maybe not the right place, but it looks like I can change the options using keyboard even when the component is disabled. We have to check it at some place.

src/components/select/bl-select.ts
muratcorlu: I also noticed this but somehow found it logical too. I think browser is keeping focused element in screen by default and if it's in a scrollable area, scrolls the container to focused element. Select options can be a long list and if we try to avoid this behaviour, we would need to handle scrolling the container to focused element if it's outside of the visible area, manually. So I'm not sure if it worths to do that.

> Also, after pressing the Space it scrolls me down of the page, I think we should preventDefault there too.

I am going to check this one.

src/components/select/bl-select.ts
muratcorlu: Hmm, good point. I'll check the native select if it has a behavior like this.

src/components/select/bl-select.ts
muratcorlu: Good catch, I'll fix it.

AykutSarac:
Any ETA on it for release?

src/components/select/bl-select.ts
muratcorlu: Fixed it.

src/components/select/bl-select.ts
muratcorlu: `ArrowDown` is also opening select from now on as in native select input.

src/components/select/bl-select.ts
muratcorlu: Fixed the scrolling issue. Preventing the default behavior helped.

scripts/generate-react-exports.js
```diff
@@ -40,29 +53,29 @@ for (const module of customElementsModules) {
...
+        react: React,
+        tagName: '${fileName}',
+        elementClass: customElements.get('${fileName}'),
+        events: ${JSON.stringify(eventNames)}
+      }
+    )
+  }))`;
+
+
+  baklavaReactFileParts.push(`export const ${componentName} = React.lazy<ReactWebComponent<${Type}${eventTypes}>>(() => ${componentDefinition(`<${Type}>`)} );`);
```
AykutSarac, line 78:
```suggestion
const componentType = `<${Type}${eventTypes}>`;
const componentPromise = `() => ${componentDefinition(componentType)}`;
const componentLazy = `React.lazy<ReactWebComponent<${componentType}>>(${componentPromise})`;
const componentExport = `export const ${componentName} = ${componentLazy};`;

baklavaReactFileParts.push(componentExport);
```

AykutSarac (approved):

buseselvi (approved):

AykutSarac:
@muratcorlu It didn't solve the value is not being updated internally, selected label is still empty. The thing is now `required` attribute now works but it doesn't check if there's a value set.

image
image
image

PR #448, opened 2023-02-24
chore: update lit

muratcorlu:
Updates Lit version to v2.6


leventozen (approved):
I don't see any breaking changes. Good work! Lots of good [features](https://github.com/lit/lit/blob/main/packages/lit/CHANGELOG.md).

Issue #328, opened 2022-12-01
Allow events without payloads


muratcorlu:
We are using our internal `@event` decorator to fire events in our components. This decorator currently doesn't allow to fire events without payloads. But in some cases (like click of button) we don't have any extra payload to send with event. Then [we write some texts as payload](https://github.com/Trendyol/baklava/blob/53867784ea3b5b1ec6b2677ea4e031de2883e276/src/components/button/bl-button.ts#L140) as a workaround. Instead we should able to fire an event without a value parameters, just like `this.onClick()`;

PR #453, opened 2023-02-28
feat: popover component

### LINKED ISSUE #323: Popover Component

ozkersemih:
Closes #323


src/components/popover/bl-popover.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  render(): TemplateResult {
+    const classes = classMap({
+      popover: true,
+      visible: this._visible,
+    });
+
+    return html`<div class=${classes}>
+      <slot id="popover" role="popover" aria-live=${this._visible ? 'polite' : 'off'}></slot>
```
muratcorlu, line 190: Apparently [there is no ARIA role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles) named `popover`. I didn't make a detailed investigation about proper A11Y guidance over popover elements, but we need to check the best practices here.

src/components/popover/bl-popover.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+  }
+
+  /**
+   * Shows popover
+   */
+  show() {
+    this._visible = true;
+    this.setPopover();
+    this.onBlPopoverShow('Show event fired!');
```
muratcorlu, line 165: No need to write that fancy message. Hopefully we can get rid of that mandatory payloads once we fix #328 

```suggestion
    this.onBlPopoverShow('');
```

src/components/popover/bl-popover.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  /**
+   * Hides popover
+   */
+  hide() {
+    this._visible = false;
+    this.onBlPopoverHide('Hide event fired!');
```
muratcorlu, line 173: No need to write that fancy message. Hopefully we can get rid of that mandatory payloads once we fix #328 

```suggestion
    this.onBlPopoverHide('');
```

src/components/popover/bl-popover.ts
```diff
@@ -0,0 +1,200 @@
...
+  @property()
+  get target(): string | Element {
+    return this._target;
+  }
+
+  set target(value: string | Element) {
+    if (typeof value === 'string') {
+      this._target = document.getElementById(value) as Element;
+    } else {
+      this._target = value;
```
muratcorlu, line 155: In theory, this setter should always get a string or an Element instance but in the wild, I think we'll hit some edge cases (like having a json object or a number, etc) and in that case it will be not very fancy to debug this issue. To avoid this, I would suggest to be more defensive here.

```suggestion
    } else if (value instanceof Element) {
      this._target = value;
    } else {
	    console.warn('BlPopover target only accepts an Element instance or a string id of a DOM element.');
```

src/components/popover/bl-popover.ts
```diff
@@ -0,0 +1,200 @@
...
+          apply(args: MiddlewareState) {
+            if(args.elements.floating && args.elements.reference){
+              Object.assign(args.elements.floating.style, {
+                width: `${args.elements.reference.getBoundingClientRect().width}px`,
+              });
+            }
+          },
+        })
+      );
+    }
```
muratcorlu, line 104: Because of those parameters (`shift`, `fitSize`) can be changed during time, setting them in `firstUpdated` will not cover all use cases. I would move all of those middleware definitions to a method that returns the list of middlewares according to current parameters, and would call it inside setPopover method.

src/components/popover/bl-popover.ts
```diff
@@ -0,0 +1,200 @@
...
+  @state() _target: string | Element;
+
+  @property({ type: Boolean })
+  shift = false;
+
+  @property({ type: Boolean, attribute: 'fit-size' })
+  fitSize = false;
+
+  @property({ type: Number })
+  offset = 10;
```
muratcorlu, line 62: Add jsdoc documents for those properties.

src/components/popover/bl-popover.stories.mdx
```diff
@@ -0,0 +1,347 @@
...
+    }
+  }}
+/>
+
+
+
+
+
+export const IconTriggerTemplate = (args) => html`
+<bl-button slot="tooltip-trigger" id="button" @click="${() => handleButtonClick("popover")}" >Trigger Popover</bl-button>
```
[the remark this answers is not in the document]
ozkersemih: Actually, I forgot to remove unused template like `IconTriggerTemplate`. We use another function that called as StoryTemplate. I will change `click` to` bl-click` on that template.

src/components/popover/bl-popover.ts
ozkersemih: So, should I remove role attribute temporarily?

src/components/popover/bl-popover.ts
ozkersemih: What does jsdoc mean? Did you mean stories?

src/components/popover/bl-popover.ts
muratcorlu: Yes, I think so. But better would be to find a proper solution.

src/components/popover/bl-popover.css
```diff
@@ -0,0 +1,31 @@
+.popover {
+  --arrow-display: var(--bl-popover-arrow-display, none);
+  --background-color: var(--bl-popover-background-color, var(--bl-color-primary-background));
+  --border-color: var(--bl-popover-border-color, var(--bl-color-primary-hover));
+  --padding: var(--bl-popover-padding, var(--bl-size-m));
+  --border-radius: var(--bl-popover-border-radius,  var(--bl-size-3xs));
+  --position: var(--bl-popover-position, absolute);
```
muratcorlu, line 7: Default value should be `fixed`, right?

src/components/popover/bl-popover.stories.mdx
```diff
@@ -0,0 +1,301 @@
...
+         args={{ id:'3', style: '--bl-popover-arrow-display: block;--bl-popover-background-color:var(--bl-color-content-tertiary);',
+            buttonText:'Popover with arrow' }}
+         play={popoverOpener}>
+    {StoryTemplate.bind({})}
+  </Story>
+</Canvas>
+
+
+## Background Color
+You can give `--bl-popover-background-color` for popover background. Default color is ` #fff`.
```
muratcorlu, line 122: Default is not `#fff`, it's `--bl-color-primary-background`

src/components/popover/bl-popover.stories.mdx
```diff
@@ -0,0 +1,301 @@
...
+         args={{  id:'4', style: '--bl-popover-background-color:var(--bl-color-featured-background);',
+         buttonText:'Custom Background'}}
+         play={popoverOpener}>
+    {StoryTemplate.bind({})}
+  </Story>
+</Canvas>
+
+
+## Border Color
+You can give `--bl-popover-border-color` for popover border color. Default color is `#ef6114`.
```
muratcorlu, line 135: Please use variables names for default values.

src/components/popover/bl-popover.stories.mdx
```diff
@@ -0,0 +1,301 @@
...
+  </Story>
+  <Story name="Custom Offset"
+         args={{  id:'10',placement:'right', offset:'40', buttonText:'Custom Offset'}}
+         play={popoverOpener}>
+    {StoryTemplate.bind({})}
+  </Story>
+</Canvas>
+
+## FitSize
+You can give `fitSize` for auto size. Default fitSize is `false`.
```
muratcorlu, line 189: Please add some hints about meaning of "auto size". You may say something like
> By default popover will have width according to its content. If you set `fitSize` as `true`, popover width will be at least as wide as the target element. This is helpful for use-cases like dropdown buttons.

src/components/popover/bl-popover.stories.mdx
```diff
@@ -0,0 +1,301 @@
...
+<Canvas>
+  <Story name="Border Radius"
+         args={{  id:'8', style: '--bl-popover-border-radius:0;', buttonText:'Border Radius 0' }}
+         play={popoverOpener}>
+    {StoryTemplate.bind({})}
+  </Story>
+</Canvas>
+
+## Offset
+You can give `offset` for padding from target. Default padding is `10`.
```
muratcorlu, line 173:
```suggestion
You can set the distance of the popover to the target element with `offset`. The default distance is `10px`.
```

By the way, are we sure if this default distance is set according to our design system values? I don't think that we have a sizing step as 10px.

@buseselvi what should be the default distance of popover element to its target element?

src/components/popover/bl-popover.stories.mdx
```diff
@@ -0,0 +1,301 @@
...
+* `left-start`
+* `left`
+* `left-end`
+* `right-start`
+* `right`
+* `right-end`
+
+If there is not enough room for the popover in the given placement, the popover is repositioned automatically according to the empty area.
+
+For example, if there is not enough room on the top, the popover is shown on the bottom.
```
muratcorlu, line 234: This relies on `shift` parameter. Maybe you can move this example to `shift` title.

src/components/popover/bl-popover.ts
```diff
@@ -0,0 +1,242 @@
...
+  | 'left-end'
+  | 'right-start'
+  | 'right'
+  | 'right-end';
+
+/**
+ * @tag bl-popover
+ * @summary Baklava Popover component
+ *
+  * @cssproperty --bl-popover-arrow-display - Sets the display of popovers arrow. Default value is `none`
```
muratcorlu, line 36:
```suggestion
 * @cssproperty --bl-popover-arrow-display - Sets the display of popovers arrow. Default value is `none`
```

src/components/popover/bl-popover.ts
```diff
@@ -0,0 +1,200 @@
...
+            const arrowDirection = arrowFlipDirections[popoverPlacement];
+
+            this.arrow.style.setProperty(arrowDirection, '-4px');
+          }
+        });
+      });
+    }
+  }
+
+  @property()
```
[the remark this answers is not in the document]
muratcorlu: Since documentation will not have separate rows for get and set, jsdoc should give information about `target` property. Like

> Sets the target element of the popover to align and trigger. It can be a string id of the target element or can be a direct Element reference of it.

src/components/popover/bl-popover.stories.mdx
buseselvi: It has 8px padding

src/components/popover/bl-popover.css
ozkersemih: If I use fixed there are some issues e.g scrolling with screen etc. I think if we use absolute it is ok. Do we have any problem with absolute in another components?

muratcorlu: OK, lets fix other issues, maybe we can try out with different values to understand the difference.

src/components/popover/bl-popover.ts
```diff
@@ -0,0 +1,270 @@
...
+  /**
+   * Target elements state
+   */
+  @state() _target: string | Element;
+
+  /**
+   * Shifts the popover along the axis
+   */
+  @property({ type: Boolean })
+  shift = false;
```
muratcorlu, line 68: This parameter enables the mode for keeping popover inside the viewport. But I'm hesitant about the need for having this property. In all of our components, this will be always `true` and I can not find a use-case to use this as `false`. Setting default value to `true` is also problemmatice since it's not technically possible to set a boolean attribute to `false` in HTML. That means if we set default value to `true` and if someone wants to set it to `false`, setting property via JS will be the only option.

buseselvi (approved):

muratcorlu (approved):

PR #458, opened 2023-03-01
feat(tab): accessibility improvements

### LINKED ISSUE #289: Accessibility improvements on Tab component
- [ ] Tab component should be usable with keyboard navigation.
- [ ] Tab component should use related `aria-` attributes. [More info](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role)

pratikgaloria:
closes #289


pratikgaloria:
@buseselvi @muratcorlu 
since we now allow tab navigation using arrow keys with this PR, the tabs will get focus and it looks like below:

image

It looks untidy at the moment, can we have a design for focused tab state? I will implement the same and then mark this PR for review.

src/components/tab-group/tab-panel/bl-tab-panel.ts
```diff
@@ -40,7 +40,7 @@ export default class BlTabPanel extends LitElement {
    * This attribute set by `tab-group` to make panel visible or hidden.
    */
   @property({ type: Boolean, reflect: true })
-  visible = false;
+  hidden = true;
```
pratikgaloria, line 43: @muratcorlu `hidden` property is more suitable here according to a11y standards, more info: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#example

buseselvi:
We added focused state design to our tab design guide 🚀 https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=179%3A2457&t=4ySnjVB8sAl5TihT-1 @pratikgaloria

src/components/tab-group/tab-panel/bl-tab-panel.ts
muratcorlu: Looks good to me.

muratcorlu:
`chore` commits don't trigger a release. I think this PR can be merged as a `feat(tab)`. Please update PR title accordingly.

src/components/tab-group/tab/bl-tab.ts
```diff
@@ -142,13 +150,15 @@ export default class BlTab extends LitElement {
...
         @click="${() => this.select()}"
+        aria-selected="${this.selected}"
       >
         <div class="title-container">
           <div class="title">${icon} ${title} ${badge}</div>
           ${caption}
         </div>
         ${helpTooltip}
       </button>
+      <span class="divider"></span>
```
muratcorlu, line 161: What was the motivation behing this decision? Did you have issues about putting border with pseude-element and outline?

pratikgaloria: earlier the design demanded to have a padding on outline from left side but not the right side, which was harder to achieve with the pseudo classes, but since we have now changed the design to have even paddings from both the sides, let me try to remove this redundant `<span>` again and check.

pratikgaloria: removed it.

src/components/tab-group/tab-panel/bl-tab-panel.ts
muratcorlu, line 43: I just realized that, setting default value of  a boolean attribute as `true` is a bit problemmatic. Because it's not technically possible to set this to false via HTML.

```html
<bl-tab-panel hidden="false">
```

doesn't work.  On the other hand, this property is supposed to set by `BlTabGroup` component only. I'm not sure if default value should be `true` or `false`.

src/components/tab-group/tab/bl-tab.css
```diff
@@ -41,6 +42,16 @@
   border-right: none;
 }
 
+:host(:focus-visible) {
+  outline: none;
+}
+
+:host(:focus-visible) .container, .container:focus-visible {
+  outline: 2px solid var(--bl-color-primary);
+  outline-offset: -3px;
```
muratcorlu, line 51: According to our sizing standards, 3px looks like a wrong value. 🤔

pratikgaloria: changed it to 4px (size-3xs).

src/components/tab-group/tab-panel/bl-tab-panel.ts
pratikgaloria: As discussed, changed this to a state instead of a property.

buseselvi:
I couldn't see the focused state screenshots on Chromatic, can we add it?

buseselvi (approved):

src/components/switch/bl-switch.stories.mdx
```diff
@@ -93,6 +99,16 @@ Disabled state can be set via `disabled` attribute. A switch can be `disabled` a
...
 </Canvas>
 
+## Keyboard navigation
+
+Switch component handles keyboard navigation and highlights itself while navigating with keyboard. First `Tab` focuses the element and user can turn on and turn off the switch with `Enter` or `Space` keys.
+
+<Canvas>
+  <Story name="Keyboard navigation" parameters={{ extraPaddingForChromatic: true }} play={focusSwitch}>
+    {SwitchTemplate.bind({})}
+  </Story>
```
muratcorlu, line 109: I think this doesn't work now. In chromatic screenshot, I don't see focus state.

muratcorlu (approved):

PR #459, opened 2023-03-02
fix(input): autofill styling

AykutSarac:
Styling issues with autofill input resolved

Screenshot 2023-03-02 at 10 31 58
Screenshot 2023-03-02 at 10 32 12
Screenshot 2023-03-02 at 10 49 10
Screenshot 2023-03-02 at 10 40 32

### DIFF
--- a/src/components/input/bl-input.css
+++ b/src/components/input/bl-input.css
@@ -45,11 +45,30 @@
   --icon-size: var(--bl-font-body-text-3-line-height);
 }
 
-.wrapper:has(input:autofill) {
+.input-wrapper {
+  --border-size: 1px;
+
+  display: flex;
+  box-sizing: border-box;
+  gap: var(--padding-vertical);
+  height: var(--height);
+  border: solid var(--border-size) var(--border-color);
+  padding: 0 calc(var(--padding-horizontal) - var(--border-size));
+  border-radius: 4px;
+}
+
+:host([disabled]) .input-wrapper {
+  cursor: not-allowed;
+  background-color: var(--bl-color-secondary-background);
+
+  --text-color: var(--bl-color-content-passive);
+}
+
+.input-wrapper:has(input:autofill) {
   background-color: var(--autofill-bg-color);
 }
 
-.wrapper:has(input:-webkit-autofill) {
+.input-wrapper:has(input:-webkit-autofill) {
   background-color: var(--autofill-bg-color);
 }
 
@@ -67,18 +86,6 @@ label {
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
-}
-
-.input-wrapper {
-  --border-size: 1px;
-
-  display: flex;
-  box-sizing: border-box;
-  gap: var(--padding-vertical);
-  height: var(--height);
-  border: solid var(--border-size) var(--border-color);
-  padding: 0 calc(var(--padding-horizontal) - var(--border-size));
-  border-radius: 4px;
 }
 
 input {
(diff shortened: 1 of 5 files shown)


src/components/input/bl-input.css
```diff
@@ -214,6 +214,10 @@ bl-icon[name='eye_on'] {
   padding: 0;
 }
 
+:host(:not([label-fixed])) :where( .input-wrapper:has(input:autofill) .wrapper:focus-within, .wrapper.has-value) label {
+  background: linear-gradient(to top, var(--autofill-bg-color) 50%, transparent 50%);
+}
```
muratcorlu, line 219: What are we doing here?

AykutSarac: It adds background to bottom half of the input label to prevent displaying white background shown here:


Makes it:

muratcorlu: I'm not sure if we need this. Because by design it has white background in disabled state as well. Maybe @buseselvi can comment about it.

buseselvi: yes, white background should stay. we dont need the other one 🙏🏻 @muratcorlu @AykutSarac

AykutSarac: Honestly it looks very odd and disturbing as an user, but if you believe it effects readability I can remove. When I checked MUI they also fill the background.

Screenshot 2023-03-02 at 11 51 24

src/components/input/bl-input.css
```diff
@@ -45,11 +45,30 @@
...
+  gap: var(--padding-vertical);
+  height: var(--height);
+  border: solid var(--border-size) var(--border-color);
+  padding: 0 calc(var(--padding-horizontal) - var(--border-size));
+  border-radius: 4px;
+}
+
+:host([disabled]) .input-wrapper {
+  cursor: not-allowed;
+  background-color: var(--bl-color-secondary-background);
```
muratcorlu, line 62: I think it's now wise to have a `--input-background-color` variable. Like:

```css
.wrapper {
   --input-background-color: var(--bl-color-content-contrast);
  ....
}

.input-wrapper {
  background-color: var(--input-background-color):
  ...
}

.wrapper:has(input:autofill) {
  --background-color: var(--autofill-bg-color);
}
```

Then you can also use `--input-background-color` for half of the floating label in case we decide to use in that way.

src/components/input/bl-input.css
```diff
@@ -45,11 +45,30 @@
...
+.input-wrapper {
+  --border-size: 1px;
+
+  display: flex;
+  box-sizing: border-box;
+  gap: var(--padding-vertical);
+  height: var(--height);
+  border: solid var(--border-size) var(--border-color);
+  padding: 0 calc(var(--padding-horizontal) - var(--border-size));
+  border-radius: 4px;
```
muratcorlu, line 57: While you touching here, can you use our border radius variable here.

src/components/input/bl-input.css
AykutSarac: I noticed that input doesn't have a background and rest of the components probably too, I think this should be handled with the same separate work

image

src/components/input/bl-input.css
AykutSarac: I didn't get what is meant here, input doesn't have a background as I mentioned at other comment.

muratcorlu: Let's consider fixing this in #462

src/components/input/bl-input.css
muratcorlu: I created a separate issue for this #462 If labels will have a change to be transparent, we can also consider it in a separate task. @buseselvi

muratcorlu (approved):

buseselvi:
We talked to the design team, the white background in the inline label prevents us from using this variant with a different background color. So we agree on removing the white background as @AykutSarac suggested.

muratcorlu:
> We talked to the design team, the white background in the inline label prevents us from using this variant with a different background color. So we agree on removing the white background as @AykutSarac suggested.

@buseselvi @AykutSarac I opened #469 for tracking this issue.

@buseselvi If you are ok with current change, just about autofill style, can you approve this PR?

buseselvi (approved):

PR #460, opened 2023-03-02
fix(input): fix listening value property changes

muratcorlu:
We realized that our input and textarea components don't listen to changes on value **property** to update their validation states. This PR solves this issue.

While solving the issue, some tests of the pagination component failed. Then I noticed some problems with how we fire and listen to `bl-change` events of pagination. So this PR also includes some refactor in the pagination component.

### DIFF
--- a/src/components/input/bl-input.ts
+++ b/src/components/input/bl-input.ts
@@ -1,4 +1,4 @@
-import { CSSResultGroup, html, LitElement, TemplateResult } from 'lit';
+import { CSSResultGroup, html, LitElement, TemplateResult, PropertyValues } from 'lit';
 import { customElement, property, query, state } from 'lit/decorators.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
@@ -178,7 +178,7 @@ export default class BlInput extends FormControlMixin(LitElement) {
 
   @state() private passwordInput = false;
 
-  private textVisiblityToggle() {
+  private textVisibilityToggle() {
     this.passwordVisible = !this.passwordVisible;
   }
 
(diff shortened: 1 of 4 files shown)


src/components/input/bl-input.ts
```diff
@@ -191,22 +191,20 @@ export default class BlInput extends FormControlMixin(LitElement) {
...
 
-  valueChangedCallback(value: string): void {
-    this.value = value;
-  }
-
   private inputHandler(event: Event) {
     const value = (event.target as HTMLInputElement).value;
 
-    this.setValue(value);
+    console.log('inputHandler', value);
```
AykutSarac, line 197: I think it's supposed to be removed?

src/components/pagination/bl-pagination.test.ts
```diff
@@ -221,7 +221,7 @@ describe('bl-pagination', () => {
         jumper.value = '5';
       }
 
-      setTimeout(() => jumper?.dispatchEvent(new Event('change')));
+      setTimeout(() => jumper?.dispatchEvent(new Event('change', { bubbles: true })));
```
AykutSarac, line 224: Move this line under of `jumper.value`, this will likely to solve the test error.

AykutSarac (changes requested):

src/components/pagination/bl-pagination.test.ts
muratcorlu: Do you mean like this?

```ts
      if (jumper) {
        jumper.value = '5';

        setTimeout(() => jumper.dispatchEvent(new Event('change', { bubbles: true })));
      }
```

This doesn't change anything.

muratcorlu: Indeed, the way we fire change event in pagination was a bit problemmatic. Also input and pagination both uses `bl-change` name and that also creates a confusion. I made some refactors in pagination and added `stopPropogation` for `bl-change` event of internal input to not fire 2 different type of `bl-change` event from pagination component. Please check latest version.

AykutSarac (approved):

pratikgaloria (approved):

Issue #405, opened 2023-02-07
Additional attributes for form field components derived from native input


muratcorlu:
There are a couple of handful attributes that we need to add to our form field components(Input, Textarea, Select, from [the native input attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#attributes).

- [x] autocomplete
- [x] autofocus
- [x] inputmode (for Input and Textarea)
- [x] `tel` and `url` for type. (for Input only)
- [x] pattern (for Input only)
- [x] spellcheck (for Textarea)

muratcorlu:
I removed `readonly` because we need to discuss about its UX since visually there is no indication about a readonly input. It seems like a bug when enabled. 

Maybe we can discuss later if we need a "readonly form element".

PR #475, opened 2023-03-13
feat: new attributes for form fields

### LINKED ISSUE #406: Dialog should focus on it's primary action button
[We realized that](#395) currently dialog component moves focus to the close button when its open. Instead we expect it to focus on primary action, if present.

### LINKED ISSUE #451: Input doesn't validate step attribute
**Describe the bug**
Input component doesn't validate number input's step attribute

Version: 2-beta.75

### LINKED ISSUE #474: Duplicate element id error inside form
**Describe the bug**
When there's more than one web-component inside a form it prints a warning in console.

image

Baklava 2.0.0-beta.80

muratcorlu:
This PR adds missing attributes defined in #405. (currently only for Input and Textarea)

Also added `autofocus` to `bl-button` to allow fixing #406 

Some remarks:

* `autofocus` doesn't work on non-native form elements on Firefox and Safari ([Source](https://caniuse.com/?search=autofocus)). So autofocus doesn't work on select component on those browsers. 
* `autocomplete` doesn't work on Select component. It needs a native form element.

Fixes #474 
Fixes #451 
Closes #406 by adding some examples in Dialog storybook document

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -70,7 +70,7 @@ const args = parseArgs(process.argv.slice(2), {
       const { host, port } = await ctx.serve(
         {
           servedir,
-          host: '0.0.0.0',
+          host: 'localhost',
         }
       );
 
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -70,7 +70,7 @@ const args = parseArgs(process.argv.slice(2), {
       const { host, port } = await ctx.serve(
         {
           servedir,
-          host: 'localhost',
+          host: '0.0.0.0',
         }
       );
 
--- a/src/components/dialog/bl-dialog.css
+++ b/src/components/dialog/bl-dialog.css
@@ -1,7 +1,9 @@
 .container {
+  --background-color: var(--bl-color-primary-background);
+
   display: flex;
   flex-direction: column;
-  background: var(--bl-color-primary-background);
+  background: var(--background-color);
   max-width: calc(100vw - var(--bl-size-4xl));
   max-height: calc(100vh - var(--bl-size-4xl));
   min-width: 424px;
@@ -21,9 +23,9 @@
   position: fixed;
 
   /* FIXME: Use z-index variable */
-  z-index: 999; 
+  z-index: 999;
  }
- 
+
 .dialog::backdrop {
   background: #273142;
   opacity: 0.7;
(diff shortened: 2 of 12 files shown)


scripts/build.js
```diff
@@ -70,7 +70,7 @@ const args = parseArgs(process.argv.slice(2), {
       const { host, port } = await ctx.serve(
         {
           servedir,
-          host: 'localhost',
+          host: '0.0.0.0',
```
leventozen, line 73: Why we change this?

src/components/dialog/bl-dialog.css
```diff
@@ -21,9 +23,9 @@
   position: fixed;
 
   /* FIXME: Use z-index variable */
-  z-index: 999; 
+  z-index: 999;
```
leventozen, line 26: We can update those after z-index variable pr is merged

scripts/build.js
muratcorlu: I needed to test playground in iOS simulator to see their behaviour in iOS, and since its `localhost` is different than my system localhost, I couldn't reach it. `0.0.0.0` means listening local IP, so I can reach playground with the IP of my computer.

After that, I used `ngrok` to test it though. Maybe I can revert this back.

src/components/dialog/bl-dialog.css
muratcorlu: This is an automatic change by my IDE, removing trailing spaces once I save a file.

leventozen: I pointed out because preventing having conflict between z-index pr and this actually. np.

scripts/build.js
leventozen: With localhost there, you can try 127.0.0.1. It should work.

muratcorlu: Yeah, maybe it works that way too. Either way, reverted this change back. 👍🏻

src/components/dialog/bl-dialog.css
muratcorlu: I'll fix the conflict, no problem.

scripts/build.js
leventozen: I mean you can try 127.0.0.1 in the simulator with our default configurations as well. So either way reverting back is better for me too.

leventozen (approved):

buseselvi (approved):

pratikgaloria (approved):

PR #482, opened 2023-03-14
feat: z-index tokens are defined

### LINKED ISSUE #305: Define z-index values
[Figma](https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=2630%3A11329)

### Discussed in https://github.com/Trendyol/baklava/discussions/276

<div type='discussions-op-text'>

<sup>Originally posted by **DamlaDemir** October  3, 2022</sup>
**Describe the bug**
The select component doesn't appear correctly.

**To Reproduce**
1. Go to https://6267ebb4dd54f5004a65f14a-xabmjaqqco.chromatic.com/?path=/docs/components-pagination--basic-usage
2. Go to select section.
3. Click show code.
4. Click select.

**Screenshots**
image

**Solution**
To fix this problem, we should increase the popover's z-index value in select. Z-index is a common problem. We should set as correctly all of our components' z-index values. We can research the best practices on this topic.
</div>

muratcorlu:
This PR defines a list of z-index tokens and replaces current hardcoded values with them.

Closes #305


buseselvi (approved):

leventozen (approved):

src/themes/default.css
```diff
@@ -54,6 +54,16 @@
   --bl-size-5xl: 4rem; /* 64px */
   --bl-size-6xl: 5rem; /* 80px */
 
+  /* Z-indexes */
+  --bl-index-deep: -1;
+  --bl-index-default: 0;
```
pratikgaloria, line 59: in the docs above, the value of default z-index is 1. we have to change that. However, I am thinking now that what is the use case of having this var? also, shouldn't the default value should be `auto`?

muratcorlu: Good catch and a valid question that I don't know the answer 😊 

Perhaps, for the time being, we should consider removing "deep" and "default" since I don't see any valid use cases for them. Furthermore, we do not currently use them in any of our components. Are you ok with that @buseselvi ?

muratcorlu: Normally we are inspired by [Nord Health Design System](https://nordhealth.design/tokens/#z-index) here but the explanation on their document is also not very convincing to me.

pratikgaloria: technically the only use-case I can think of is that you want to take the element out of its natural stack order and put it in the base, so maybe we can still keep it but change the name to `--bl-index-base` while considering our base value as 0, wdyt?

pratikgaloria: because the name `-default` can be confusing here and ppl start using it even when it's not necessary.

muratcorlu: Yeah, makes sense to me. I pushed some changes. Please check the current version. @buseselvi @leventozen

buseselvi (approved):

pratikgaloria (approved):

PR #483, opened 2023-03-15
feat(popover): max-width and border-size css properties are added

### LINKED ISSUE #478: Using popover component in Tooltip
Tooltip component should use our Popover for its content.

ozkersemih:
This PR contains:

* Tooltip starts using popover component internally.
* Popover component has 2 additional CSS properties: `--bl-popover-border-size` and `--bl-popover-max-width`
* Some optimizations and fixes done on positioning arrow in popover.
* UglifyCSS which is used internally by lit-css plugin that we use to minify our styles breaks the code by replacing all `0px` values to `0`. This was causing problem about using this value in CSS calc, and [not a safe thing at all](https://www.oddbird.net/2022/08/04/zero-units/). Since UglifyCSS project is not maintained anymore I manually replaced its functionality with clean-css library in build script. (I [created an issue](https://github.com/bennypowers/lit-css/issues/43) to replace uglifycss with something better in lit-css plugin as well)

Closes #478


src/components/tooltip/bl-tooltip.css
```diff
@@ -3,34 +3,17 @@
...
-  z-index: 999;
-  width: max-content;
-  hyphens: auto;
+.wrapper{
+  display: inline-block;
+}
 
+.tooltip {
+  --bl-popover-background-color: var(--bl-color-secondary);
+  --bl-popover-border-color: var(--bl-color-secondary);
```
muratcorlu, line 12: I think we should use content color instead.

```suggestion
  --bl-popover-background-color: var(--bl-color-content-primary);
  --bl-popover-border-color: var(--bl-color-content-primary);
```

src/components/tooltip/bl-tooltip.css
```diff
@@ -3,34 +3,17 @@
...
+  --bl-color-content-primary: white;
```
muratcorlu, line 13: Instead I would do this:

```css
color: var(--bl-color-content-primary-contrast);
```

src/components/tooltip/bl-tooltip.css
```diff
@@ -3,34 +3,17 @@
...
+  --bl-popover-arrow-display: inline-block;
```
muratcorlu, line 14: In popover documentation we mentioned this can take either `none` (default) or `block`. But now we set `inline-block`. This was one of my hesitation points about using `none` as default 😊

src/components/tooltip/bl-tooltip.test.ts
```diff
@@ -25,13 +26,13 @@ describe('bl-tooltip', () => {
        class="trigger"
        name="tooltip-trigger">
       </slot>
-      <div class='tooltip'>
+      <div class='wrapper'>
```
muratcorlu, line 29: Why do we need this wrapper?

src/components/tooltip/bl-tooltip.css
```diff
@@ -3,34 +3,17 @@
...
-  border: none;
-  background-color: var(--bl-color-secondary);
-  color: var(--bl-color-content-primary-contrast);
-  border-radius: var(--bl-size-3xs);
-  pointer-events: none;
-  font: var(--bl-font-title-3-regular);
-  padding: var(--bl-size-m);
-  z-index: 999;
-  width: max-content;
-  hyphens: auto;
```
muratcorlu, line 18: I think we still need those 2.

src/components/tooltip/bl-tooltip.test.ts
ozkersemih: I have used wrapper class to set display as inline-block. If I don't use it every tooltip displayed on single row(block). But maybe am I missing something?

src/components/tooltip/bl-tooltip.ts
```diff
@@ -44,6 +45,39 @@ export default class BlTooltip extends LitElement {
...
+      this._target = value;
+    } else {
+      console.warn(
+        'BlTooltip target only accepts an Element instance or a string id of a DOM element.'
+      );
+    }
+
+    if (this._target instanceof Element) {
+      this._target.addEventListener('focus', this.show.bind(this));
+      this._target.addEventListener('blur', this.hide.bind(this));
```
muratcorlu, line 75: Listening to focus and blur events would be problemmatic on some elements like input. We will not want to show a tooltip when user focuses to an input by clicking it. I think we need to re-consider this behaviour.

src/components/tooltip/bl-tooltip.ts
```diff
@@ -142,25 +133,29 @@ export default class BlTooltip extends LitElement {
...
+    >
+    </slot>`;
+  }
+
+  render(): TemplateResult {
+    return html` ${!this._target ? this.triggerTemplate() : ''}
+      <div class="wrapper">
+        <bl-popover
+          class="tooltip"
+          .target="${!this._target ? this.trigger : this._target}"
```
muratcorlu, line 154: We set target here and also in `show` method. I think one should be dropped. Probably we need to drop this line in favor of `show` method.

muratcorlu:
@ozkersemih After fixing failing test, I noticed some UI issues regarding to the border size addition comparing to current tooltip design. Also `max-width` of tooltip was not working properly. To keep design aligned, I added some extra css variables to popover.

src/components/tooltip/bl-tooltip.css
```diff
@@ -3,34 +3,18 @@
...
+  display: inline-block;
+  color: var(--bl-color-neutral-full);;
 }
 
-.tooltip:not(.visible) {
-  visibility: hidden;
+bl-popover {
+  --bl-popover-background-color: var(--bl-color-neutral-darker);
+  --bl-popover-arrow-display: block;
+  --bl-popover-border-size: calc(0 * 1px);
```
muratcorlu, line 14: The current implementation of popover doesn't accept unitless border-size (like `0`). And our linter doesn't accept `0px` since it's a best practice to write `0` without unit. I asked for a solution to this on stackoverflow: https://stackoverflow.com/questions/76179826/css-calc-with-a-variable-that-can-be-unitless

muratcorlu: With the answers to SO question, I come up with the idea to change our linter to accept (even suggest) using units with 0 value as this article suggests: https://www.oddbird.net/2022/08/04/zero-units/

muratcorlu (approved):

buseselvi:
Radio Group selected option seems wrong:

image
image

And I didn't understand why secondary and tertiary button hovers changed in the UI tests with this implementation, storybook looks ok (unchanged) but in chromatic, hover states look the same as the default state. @muratcorlu @ozkersemih

buseselvi (approved):

PR #503, opened 2023-03-28
feat(input): set custom error state

### LINKED ISSUE #500: CustomInvalidText reactivity bug of input component
**Describe the bug**
When we change customInvalidText of input component via react state; changes aren't apply directly dom, after that new customInvalidText apply to dom with keypress again.

**To Reproduce**
`<BlInput type={'number'} required min={15} customInvalidText={'It is required'} />`
When `1` is written in the above input, customInvalidText changes as a `'Must least 15'` with react state. 
After that input clear and customInvalidText changes as a `'It is required'` with react state.
Invalidation text at dom occur  `'Must least 15'` but should be `'It is required'` like given prop.

**Expected behavior**
Given prop as customInvalidText directly apply to dom. For example `helpText` prop runs completely sync given prop and dom.

**Screenshots**
Screen Shot 2023-03-28 at 10 10 49

**Additional context**
Related issue: https://github.com/Trendyol/baklava/issues/502

### LINKED ISSUE #502: Input component validation status
Min, max etc native validation features are not enough sometimes. Is it possible; input component validation status give from outside via prop.

`<BlInput valid={false} customInvalidText={state.changeableInvalidText}>`
Prop name may be different(valid, customValidity etc).

**Additional context**
Related issue: https://github.com/Trendyol/baklava/issues/500

muratcorlu:
* Makes `invalid-text` attribute (or `customInvalidText` property) reactive to update error message by setting the property.
* I added `forceCustomError()` method to set validation state to `invalid` for custom errors (those are not handled with basic validation rules). In this state input will be in invalid state by showing error message defined with `invalid-text`. And another method `clearCustomError()` clears this state. I'm not completely sure if those are good method names for this use cases. Any suggestions are very welcome.

Fixes #500 
Closes #502


yasin749 (approved):

src/components/input/bl-input.ts
```diff
@@ -210,8 +219,27 @@ export default class BlInput extends FormControlMixin(LitElement) {
     return this.customInvalidText || this.validationTarget?.validationMessage;
   }
 
+  /**
+   * Force to set input as in invalid state.
+   */
+  forceCustomError() {
+    this.validationTarget.setCustomValidity(this.customInvalidText || 'An error ocurred');
```
pratikgaloria, line 226: a small typo here, `occurred`

yasin749 (approved):

PR #506, opened 2023-03-29
fix(input): label style and background color fixed

### LINKED ISSUE #462: Form fields are transparent
Our Input, Textarea and Select components don't have a background-color by default and this causes inconsistent results regarding to the containers.

### LINKED ISSUE #469: Floating form labels should be transparent
Floating form labels will be transparent according to the decision done by UX team. This will need change on components those have floating labels like Input, Textarea and Select.

> We talked to the design team, the white background in the inline label prevents us from using this variant with a different background color. So we agree on removing the white background as @AykutSarac suggested.

_Originally posted by @buseselvi in https://github.com/Trendyol/baklava/issues/459#issuecomment-1457714035_

muratcorlu:
* Updates floating label of Input, Textarea and Select components according to latest change. Now labels don't have white background.
* Background color set for Input, Textarea and Select

Fixes #469 
Fixes #462

### DIFF
--- a/src/components/input/bl-input.css
+++ b/src/components/input/bl-input.css
@@ -14,10 +14,11 @@
   --icon-size: var(--line-height);
   --padding-vertical: calc((var(--height) - var(--line-height)) / 2);
   --padding-horizontal: var(--bl-size-xs);
+  --label-padding: var(--bl-size-3xs);
+  --background-color: var(--bl-color-content-primary-contrast);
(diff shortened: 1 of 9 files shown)


buseselvi:
Select input texts does not look centered @muratcorlu 
image

muratcorlu:
There are really challenging issues on this topic. Still working on fixing all of them. I'll pull this PR back to draft until I fix all of them.

DamlaDemir (approved):

PR #510, opened 2023-04-02
fix(popover): using min-width for fit-size

### LINKED ISSUE #480: Using Popover component in Dropdown Button
Dropdown component should use Popover component in it to show action list

ogunb:
This PR implements using popover component inside dropdown component by replacing local implementation. To do that we needed to change `fit-size` behaviour of popover. If `fit-size` set, popover will have the minimum size of it's trigger element. If content will be larger, than popover will increase until it's max-width.

Fixes #480

### DIFF
--- a/src/components/dropdown/bl-dropdown.ts
+++ b/src/components/dropdown/bl-dropdown.ts
@@ -1,15 +1,6 @@
 import { LitElement, html, CSSResultGroup, TemplateResult } from 'lit';
 import { customElement, property, state, query } from 'lit/decorators.js';
-import {
-  computePosition,
-  flip,
-  offset,
-  autoUpdate,
-  size,
-  MiddlewareArguments,
-} from '@floating-ui/dom';
 import { event, EventDispatcher } from '../../utilities/event';
-import { classMap } from 'lit/directives/class-map.js';
 
 import style from './bl-dropdown.css';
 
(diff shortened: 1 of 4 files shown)


src/components/dropdown/bl-dropdown.ts
```diff
@@ -199,9 +147,9 @@ export default class BlDropdown extends LitElement {
       >
         ${this.label}
       </bl-button>
-      <div class="${popoverClasses}" role="menu" aria-expanded="${this.opened}">
+      <bl-popover .target="${this}" placement="bottom-start" @bl-popover-hide="${this.close}">
```
muratcorlu, line 150: Target should be the button above, not dropdown component itself.

ogunb: When I try to pass the target as the query result of the button, the popover component logs a warning because the button is not yet rendered, and the query result is null at initial render.

I checked the tooltip implementation PR, and it looks like the target is set on the popover with a show callback. We can implement it that way, if that's ok.

muratcorlu: Maybe it's time to try [`ref` directive](https://lit.dev/docs/templates/directives/#ref) of Lit, which is built for that kind of purposes. If it doesn't help, setting this lazily either in `firstUpdated` or just before showing it would also work.

muratcorlu:
@ogunb There was an issue about chromatic tests that is also blocking pipelines of this PR. I merged the fix (#509). I merged next to this PR to use the fix. Please pull changes to your local.

src/components/dropdown/bl-dropdown.ts
muratcorlu: LGTM. Please make your PR ready for review.

muratcorlu:
Because now this PR is not just a refactor, and changes behaviour of popover component, I updated PR title and description accordingly.

muratcorlu (approved):

PR #526, opened 2023-04-12
chore(deps): support node v18

### LINKED ISSUE #361: Node 18 support

AykutSarac:
It worked fine on my end but I recommend testing locally.
Closes #361


muratcorlu:
I know why the pipeline fails: #173 We need to find the proper way of accepting PRs from forks.

muratcorlu:
As far as I see, [storybook v6 doesn’t support node v18](https://github.com/storybookjs/storybook/issues/16555) without [disabling new openssl provider](https://github.com/storybookjs/storybook/issues/16555#issuecomment-1156436628) which is not a great solution.

I think we should consider fixing this together with #522

muratcorlu:
Yeah, I think upgrading to v7 will need a bigger re-organization of our storybook files. Let's connect on that.

muratcorlu:
With MDX2, [they deprecated writing inline stories](https://storybook.js.org/docs/react/migration-guide#storiesmdx-to-mdxcsf) in MDX file. But maybe we can keep them for while if it works. I'll have a look at the branch.

muratcorlu:
I decided to separate tasks by first fixing storybook upgrade (which already works well with both Node 16 and 18) then returning back to node 18 issue. Let's first merge #540 and return back to this. @AykutSarac 

PS: Once we got everything running in node 18 on local, we also need to check our pipeline config, because some of our jobs are forced to use node v16 now.

AykutSarac:
@muratcorlu what do you think about replacing npm with yarn in this PR?

muratcorlu (approved):

Issue #525, opened 2023-04-12
Using icon component with custom icon sets


muratcorlu:
I think it would be nice to allow people to use our icon component with custom, external icon sets. I think [we can inspire from Shoelace](https://shoelace.style/components/icon?id=icon-libraries) for this and give the ability to register custom icon sets by still keeping current icon set by default.

```html
<script type="module">
  import { registerIconLibrary } from '@trendyol/baklava';

  registerIconLibrary('bootstrap', {
    resolver: name => `https://cdn.jsdelivr.net/npm/bootstrap-icons@1.0.0/icons/${name}.svg`
  });
</script>

<bl-icon library="bootstrap" name="arrow-up"></bl-icon>
```

ogunb:
We have decided against implementing this feature for the time being.

Users can change the icon pack by using the `setIconPath` utility, provided they supply all icons with the same names. However, we do not endorse or officially support this method.

We may revisit this feature in the future. If anyone feels the need for this functionality, please feel free to reopen the issue.

PR #531, opened 2023-04-17
feat(icon): add type support for icon name

AykutSarac:
This PR adds type support for icon name, this will bring type support for projects using TypeScript and help us potentially avoid errors in future.


muratcorlu:
To avoid breaking changes, we need to find a way to work this with multiple icon libraries which we want to have soon: #525

AykutSarac:
@muratcorlu Is there a plan to seperate icon lib to a new package or can we bring type support for icons?

muratcorlu:
Yes, we are planning to exclude icons to a separate package: https://github.com/trendyol/baklava-icons

muratcorlu:
Closing this for now in favor of #656

PR #532, opened 2023-04-17
fix: fix react wrapper types

muratcorlu:
This PR fixes missing types in React wrappers.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,1 @@
 engine-strict=true
-legacy-peer-deps=true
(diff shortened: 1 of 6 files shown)


AykutSarac (changes requested):
At the generated d.ts file `ISelectOption` interface requires a parameter so we need to pass a default type:
```ts
// src/components/select/bl-select.ts
interface ISelectOption<T = string>
```

Another issue is `ValueType` is not defined here:
```ts
// baklava-react.d.ts
export declare const BlSelectOption: React.LazyExoticComponent<ReactWebComponent<BlSelectOptionType<string>, {
    onBlSelectOption: EventName<CustomEvent<ValueType | string | null>>;
    onBlFocus: EventName<CustomEvent<ValueType | string | null>>;
    onBlBlur: EventName<CustomEvent<ValueType | string | null>>;
}>>;
```

Issue #529, opened 2023-04-14
Dropdown should be closed when an item is clicked


muratcorlu:
Currently dropdown keeps open when an action in its popover is clicked. I think general expectation is that it should automatically closed whenever a dropdown action is selected.

PR #533, opened 2023-04-17
fix(dropdown): close dropdown when an item is clicked

abugraokkali:
This pull request adds automatic closing of the dropdown menu when an action is selected in the popover. This improves the user experience by eliminating the need for users to manually close the menu.

Fixes https://github.com/Trendyol/baklava/issues/529

### DIFF
--- a/src/components/dropdown/bl-dropdown.test.ts
+++ b/src/components/dropdown/bl-dropdown.test.ts
@@ -91,6 +91,32 @@ describe('bl-dropdown', () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


muratcorlu:
Hi @abugraokkali Thanks for the contribution. I’ll check the issue about the failing pipeline.

muratcorlu (approved):
Looks very good to me! Thanks for this great contribution @abugraokkali !

I'll just consider this as a `fix` instead of `feat`, since I think this should be the expected behavior already.

Issue #543, opened 2023-04-21
Closing Popover with Escape should not cause exiting browser fullscreen


muratcorlu:
In Safari, with full-screen, if you close a popover with Escape button, browser exits from fullscreen mode. We should prevent this.

Similar to #542

PR #551, opened 2023-04-25
fix(popover): prevent exiting fullscreen with popover close

abugraokkali:
Prevented the browser from exiting fullscreen mode when the popover is closed with the Escape key. 
This was accomplished by using the keydown event instead of the keyup event, and making sure the action is only triggered when the Escape key is pressed down.

Fixes https://github.com/Trendyol/baklava/issues/543


muratcorlu:
Thank you @abugraokkali ! 💯

Issue #561, opened 2023-04-27
@bl-dialog-close event is firing without opening the dialog


mahmudmardini:
**Describe the bug**
When we add a boolean variable to open prop and this variable is false, the @bl-dialog-close event is fired automatically.
That means if we have a logic to execute when a user close the dialog, this logic will be executed in default even if the user never opened the dialog.

Example:
<bl-dialog
  :open="isShowModal = false"
  @bl-dialog-close="onClose"
>
content
</bl-dialog>

**To Reproduce**
1. Go to <bl-dialog> component
2. Add a boolean variable that is false in default for 'open' property
3. Add a test method (onClose) that executes when @bl-dialog-close event is fired
4. The onClose method will be executed without opening the dialog.

**Expected behavior**
The onClose method shouldn't be executed until the user clicks on the dialog close button.

mahmudmardini:
I'm closing this issue since it's fixed with #594.

PR #562, opened 2023-04-28
fix(dialog): fire dialog close / open events by checking current open status

mahmudmardini:
This pull request prevents firing bl-dialog-close event automatically when the component is created with open=false property.
Fixes: #561

### DIFF
--- a/src/components/dialog/bl-dialog.test.ts
+++ b/src/components/dialog/bl-dialog.test.ts
@@ -303,7 +303,7 @@ describe('bl-dialog', () => {
       });
 
       it('should not fire bl-dialog-close event when dialog open prop is false in default', async () => {
-        const el = await fixture<typeOfBlDialog>(html`<bl-dialog :open="false" caption="My title">
+        const el = await fixture<typeOfBlDialog>(html`<bl-dialog .open=${false} caption="My title">
         </bl-dialog>`);
 
         setTimeout(async () => {
--- a/src/components/dialog/bl-dialog.test.ts
+++ b/src/components/dialog/bl-dialog.test.ts
@@ -301,6 +301,16 @@ describe('bl-dialog', () => {
           expect(ev.detail.isOpen).to.equal(false);
         });
       });
+
+      it('should not fire bl-dialog-close event when dialog open prop is false in default', async () => {
+        const el = await fixture<typeOfBlDialog>(html`<bl-dialog :open="false" caption="My title">
+        </bl-dialog>`);
+
+        setTimeout(async () => {
+          const ev = await oneEvent(el, 'bl-dialog-close');
+          expect(ev).to.not.exist;
+        });
+      });
     });
   });
 });
--- a/src/components/dialog/bl-dialog.ts
+++ b/src/components/dialog/bl-dialog.ts
@@ -54,7 +54,7 @@ export default class BlDialog extends LitElement {
   @event('bl-dialog-close') private onClose: EventDispatcher<object>;
 
   updated(changedProperties: PropertyValues<this>) {
-    if (changedProperties.has('open') && changedProperties.get('open') !== undefined || this.open){
+    if (changedProperties.has('open') && changedProperties.get('open') !== undefined || this.open) {
       this.toggleDialogHandler();
     }
   }
@@ -75,7 +75,7 @@ export default class BlDialog extends LitElement {
       this.toggleFooterShadow();
       window?.addEventListener('keydown', event => this.onKeydown(event));
       window?.addEventListener('resize', () => this.toggleFooterShadow());
-    } else if(!this.open) {
+    } else {
       this.dialog?.close?.();
       this.onClose({ isOpen: false });
       document.body.style.overflow = 'auto';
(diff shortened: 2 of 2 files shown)


mahmudmardini:
This pull request prevents firing bl-dialog-close event automatically when the component is created with open=false property.
Fixes: #561

muratcorlu:
> This pull request prevents firing bl-dialog-close event automatically when the component is created with open=false property.
> Fixes: #561

Please move this comment to PR description.

src/components/dialog/bl-dialog.test.ts
```diff
@@ -301,6 +301,16 @@ describe('bl-dialog', () => {
           expect(ev.detail.isOpen).to.equal(false);
         });
       });
+
+      it('should not fire bl-dialog-close event when dialog open prop is false in default', async () => {
+        const el = await fixture<typeOfBlDialog>(html`<bl-dialog :open="false" caption="My title">
```
muratcorlu, line 306: `:open` doesn’t mean same with `lit-html` as `vuejs`. Basically here you are setting a property named `:open` with value of string `false`.  Actually there is no way to set a boolean attribute to `false` in HTML because attribute values are always string. As “boolean attribute” we mean just existence of an attribute. So `<bl-dialog open>` means a `true` value for `open`. If you don’t set `open`, it’ll be false. So `open=“false”` still means a truety value for `open` property. But you can set property instead of attribute as `.open=${false}`. This sets the property of input class  with a real `false` value, instead of attribute.

src/components/dialog/bl-dialog.ts
```diff
@@ -55,7 +55,7 @@ export default class BlDialog extends LitElement {
 
   updated(changedProperties: PropertyValues<this>) {
     if (changedProperties.has('open')) {
```
muratcorlu, line 57: I think we can just check if it’s really changed here to take an action:
```suggestion
    if (changedProperties.has('open') && changedProperties.get('open') !== this.open) {
```

In this case we would not need to further checks in `toggleHandler`.

mahmudmardini: I tried this solution before but unfortunately, it didn't solve the close-event firing issue

olkeoguz: Have you tried the following condition? 

` if (changedProperties.has('open') && changedProperties.get('open') !== undefined) `

I think the reason why @muratcorlu's suggestion doesn't work is that the property is assigned as undefined in phase of component getting attached to the document's DOM.

DamlaDemir: You are right @olkeoguz but we missed one thing according to your solution first open event won't trigger

You can try this:

```suggestion
    if (changedProperties.has('open') && (changedProperties.get('open') !== undefined || this.open)) {
```

src/components/dialog/bl-dialog.test.ts
mahmudmardini: Thanks a lot for this information. 
I'm new to web components and this is my very first PR in Baklava 🙏

src/components/dialog/bl-dialog.ts
mahmudmardini: Yes that fixed the issue, thanks all

muratcorlu: Please make sure this prevents running toggle handler multiple time in case of setting same value again, like:

```js
const dialog = document.querySelector(‘bl-dialog’);

dialog.open = true;

// Below should not do any change
dialog.open = true;
```

or

```js
const dialog = document.querySelector(‘bl-dialog’);

dialog.open = false;

// Below should not do any change
dialog.open = false;
```

src/components/dialog/bl-dialog.ts
```diff
@@ -54,7 +54,7 @@ export default class BlDialog extends LitElement {
   @event('bl-dialog-close') private onClose: EventDispatcher<object>;
 
   updated(changedProperties: PropertyValues<this>) {
-    if (changedProperties.has('open')) {
+    if (changedProperties.has('open') && changedProperties.get('open') !== undefined || this.open) {
```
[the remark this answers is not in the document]
mahmudmardini: Yes, that's right I forgot to add parentheses, I added them with the last commit.

src/components/dialog/bl-dialog.test.ts
```diff
@@ -301,6 +301,16 @@ describe('bl-dialog', () => {
           expect(ev.detail.isOpen).to.equal(false);
         });
       });
+
+      it('should not fire bl-dialog-close event when dialog open prop is false in default', async () => {
+        const el = await fixture<typeOfBlDialog>(html`<bl-dialog .open=${false} caption="My title">
```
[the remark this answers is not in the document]
mahmudmardini: I'm not sure how to make this test work as expected
do you have any suggestions?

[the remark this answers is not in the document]
MehmetCanBOZ: I think maybe, you can use [sinon](https://www.npmjs.com/package/sinon) library and install it as a dev dependency.

Use the **spy** function to check if the event is fired or not.
```
import { spy } from 'sinon';

it('should not fire bl-dialog-close event when dialog open prop is false in default', async () => {
  const spyEventHandler = spy();

  await fixture<typeOfBlDialog>(
    html`<bl-dialog .open=${false} @bl-dialog-close=${spyEventHandler}></bl-dialog>`
  );

  expect(spyEventHandler).to.have.callCount(0);
});
```

mahmudmardini:
I'm closing this issue since it's fixed with #594.

PR #570, opened 2023-05-08
fix: set package type as module

AykutSarac:
This pull request adds the "type: module" field to the package.json file. This change is made to indicate that our component library uses ECMAScript (ES) modules syntax and is designed to be consumed as an ES module.

Motivation:
By adding "type: module", we explicitly inform the JavaScript runtime and developers that our library follows the ES modules standard. This helps in promoting interoperability and enables seamless integration with modern JavaScript environments that natively support ES modules.

Benefits:
1. Clearer intent: By explicitly stating that our library uses ES modules, it becomes easier for developers to understand and utilize the library in projects that support ES modules.

2. Improved interoperability: With "type: module" specified, consumers of our library using tools like bundlers, module loaders, or build systems that understand ES modules can effortlessly incorporate our library into their projects.

3. Consistency: This change aligns our library with the industry trend of adopting ES modules, allowing for consistency across different projects and libraries that leverage ES modules.

Note:
Adding "type: module" implies that our library will not be compatible with older environments that only support CommonJS syntax. If we need to support both CommonJS and ES modules, we can explore using a build tool like Babel to transpile the code into CommonJS format and include both versions in the package.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -6,7 +6,8 @@ module.exports = {
   },
   extends: ['eslint:recommended', 'prettier', 'plugin:storybook/recommended'],
   env: {
-    browser: true
+    browser: true,
+    node: true
   },
   globals: {
     process: 'readonly'
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "@trendyol/baklava",
+  "type": "module",
   "version": "0.0.0-dev",
   "description": "Trendyol Baklava Design System",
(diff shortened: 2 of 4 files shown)


AykutSarac:
Do you think we can merge this with #526?

muratcorlu:
@AykutSarac Can we first merge this and check #526 with it?

muratcorlu:
Since #526 doesn't trigger a release, I merged it first. Once pipeline passes on next branch, we can merge this as well.

Issue #176, opened 2022-07-29
Determining Animation Principles For Components


DamlaDemir:
We can use animation for transitions of components such as `tooltip, tab`. At this point, we should create a generic structure for animation principles.

muratcorlu:
## Design Meeting Report about Animation Principles

Date: 21.12.2022 11:00 (CET)
---

* We'll create a list of transitions to be used in Baklava in general
* We'll create a list of durations to be used in Baklava in general. Durations will have a ratio between them. So changing the main duration in design tokens would make all of the animations slower or faster accordingly with a stable ratio
* We may create a list of animations that will have a flow and combination of transitions and durations. Like an animation named `slide-in-out` can define entering and leaving animations by different transitions and durations (example: enter from right in `slow` duration and leave to right in `fast` duration)
* Animations should respect "reduce motion" accessibility setting and disabled in that case
* Animation update requirements for the components will be considered separate tasks
* Animation requirements for upcoming components will be discussed as part of design meetings
* Since the duration list will be a list of design tokens, it'll make it possible to customize them. But we'll not give the option to disable animations for components specifically until we really see a valid requirement for this.

PR #594, opened 2023-05-24
feat(dialog): bl-dialog-request-close event added

### LINKED ISSUE #564: Dialog should not close when we click somewhere outside the dialog
Hello,
We save data with a form in the dialog. While filling out a form in the bl-dialog, the user can accidentally click outside the dialog and close the dialog.
Can it be optional to close the dialog when clicking outside the dialog ?

muratcorlu:
This PR adds a new "cancellable" event named `bl-dialog-request-close` to dialog which allows the developer to prevent closing dialog in specific cases. Event has the source info as the internal initiator of this closing request. Options are `backdrop`, `keyboard` (Escape button) and `close-button` (built-in close button on top right).

Maybe a bouncing or shaking animation can be added in the case of prevention. @buseselvi may comment about this if we have a decision already, otherwise it can be defined in an upcoming PR.

Fixes #564


buseselvi:
When we determine the animation principles ( #176 ), we can add animation to this dialog later. 🙏🏻

Enes5519:
I don't think it should be a new event. We should be able to do this via the 'bl-dialog-close' event.

muratcorlu:
> I don't think it should be a new event. We should be able to do this via the 'bl-dialog-close' event.

Nice point. I considered about the difference of one or separate events for this and noticed a strong advantage of having a separate event for this use case. Check example below:

```js
document.querySelector('bl-dialog').addEventListener('bl-dialog-close', (event) => {
  console.log('This will prevent closing the dialog');
  event.preventDefault();
});

document.querySelector('bl-dialog').addEventListener('bl-dialog-close', (event) => {
  console.log('This will still work even tough dialog is not actually closed');
});
```

Consider multiple unrelated parts listen `bl-dialog-close` events to have separate actions. If one of them prevents closing dialog, other listeners will not aware of this.

Enes5519:
I don't think this is a very real world case. But if such a scenario happens, the developer knows they blocked the event somewhere. Even though he still doesn't know, he can check it with [defaultPrevented](https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented).

muratcorlu:
`defaultPrevented` doesn't work for this use case.

```js
document.querySelector('bl-dialog').addEventListener('bl-dialog-close', (event) => {
  console.log('This will prevent closing the dialog');
  event.preventDefault();
});

document.querySelector('bl-dialog').addEventListener('bl-dialog-close', (event) => {
  if (!event.defaultPrevented) {
    console.log('This will still work even tough dialog is not actually closed');
  }
});
```

muratcorlu:
`defaultPrevented` is not safe since it relies on the order of event listeners. Check this:

```js
document.querySelector('bl-dialog').addEventListener('bl-dialog-close', (event) => {
  if (!event.defaultPrevented) {
    console.log('This will still work even tough dialog is not actually closed');
  }
});

document.querySelector('bl-dialog').addEventListener('bl-dialog-close', (event) => {
  console.log('This will prevent closing the dialog');
  event.preventDefault();
});
```

If you change the order, it works. But in a big complex codebase with a big team, it's very tricky to ensure which part will work first.

AykutSarac:
Why don't we have only `open` and `on-close` attributes for the dialog like at [Mantine](https://mantine.dev/core/dialog/)?
The component will watch for the `open` attribute to toggle the visibility and developer will provide the function for `on-close` event. For the component side, this method will be attached to the parts where the close event should be triggered such as outside click. That way developer can decide when to prevent close event or even trigger couple of other actions together.

Enes5519:
As I said, it is not something that will happen in the real world for more than one place to listen to this event. If there is no event such as 'change-before' in Native, if we can prevent it in 'change' event, it should be the same here. The developer should know what he is doing.

muratcorlu:
@AykutSarac So you mean dialog should not actually hide itself, instead just fire a close event once it's intended to close, and then developer should listen this event and call `close()` method of the dialog, right? Then, boilerplate code for a basic dialog will be a lot, since developer will "always" need to listen close event to make a simple dialog closable. I'm not sure if this will be a practical way. Another problem is that this approach will be a breaking change for the current situation, and we'll need to make a major release for that. I don't know it worth. Native HTML dialog also automatically closes by pressing "Esc" button without writing any extra code.

@Enes5519 Can you give a working example for blocking change event of a native input? I tried but couldn't manage. (https://jsfiddle.net/gterbkL2/1/) 

> it is not something that will happen in the real world for more than one place to listen to this event

> The developer should know what he is doing.

I can see valid potential use cases. There can be a generic analytics library that logs how many dialogs are opened and closed, by just listening to those events globally. Even this library can be done by a separate team. So the development teams of the two separate code that listens to the same event could be independent. 

I believe the success and stability of libraries (especially design systems) should not rely on developer quality. Of course, we can not prevent the developer from making mistakes in every case, but what stops us here if there is a safer way to avoid making mistakes?

AykutSarac:
> @AykutSarac So you mean dialog should not actually hide itself, instead just fire a close event once it's intended to close, and then developer should listen this event and call `close()` method of the dialog, right? Then, boilerplate code for a basic dialog will be a lot, since developer will "always" need to listen close event to make a simple dialog closable. I'm not sure if this will be a practical way. Another problem is that this approach will be a breaking change for the current situation, and we'll need to make a major release for that. I don't know it worth. Native HTML dialog also automatically closes by pressing "Esc" button without writing any extra code.

Yes, exactly. You should prioritize a community to improve ease of coding, in this case we have to compare how many developers you expect to use Baklava in vanilla JS and how many will use it with a framework like React/Vue. I'm currently using Baklava with React and only way to toggle dialog for me is to use ref which is not the practical way for it - especially dialog and modals -.

Currently, it is a must to watch changes for the `open` attribute to make real use of the dialog element, and manually firing the onClose event is a great idea because you can decide if it should be closed or trigger other events before closing dialog without adding a new event specifically for this purpose.

muratcorlu:
@AykutSarac We can not change the behavior of dialog without releasing v3.0. But I hear you, and your expectations are valid. But in the meantime, this new event will already allow you to always open/close dialog with `open` property, by just listening `onBlDialogBeforeClose` event and `preventDefault`. Something like:

```js
const [opened, { close }] = useDisclosure(false);

<BlDialog open={opened} onBlDialogBeforeClose={(e) => 
	e.preventDefault();
	close();
}}> ... </BlDialog>
```

And this will not be a breaking change and will cover all of our requirements. Is this not good enough?

Enes5519:
You are right. I can't prevent change or input event of input elements. I can block the keydown event, the approach here is really like the before event. I would appreciate it if you could also review the example below. [https://jsfiddle.net/9k7rbdvy/5/](https://jsfiddle.net/9k7rbdvy/5/)

I want to give you a sample code output. I have to handle it like this on my own because the value I hold in my own case is not a two-way binding.
```js
function onToggle(e: CustomEvent) {
    const isSuccess = dispatch("toggle", e.detail, { cancelable: true });
    if (isSuccess) {
      value = e.detail;
    } else {
      (e.target as HTMLElement).setAttribute("checked", `${value}`);
    }
  }
```

Enes5519 (approved):
In Native, cancelable events appear to be the event before the actual event, so doing it this way doesn't feel too wrong anymore.

AykutSarac (approved):

muratcorlu:
According to our internal discussion, I renamed event as `bl-dialog-request-close`, since this event is not fired when we programmatically close dialog by setting `open` property.

Also fixed an issue causes firing `bl-dialog-close` event when dialog first initiated. @AykutSarac @Enes5519

PR #605, opened 2023-05-26
feat(button): form property added for setting target form

### LINKED ISSUE #603: form attribute for buttons
Buttons are trying to find form element using `closest` function. I want to specify which form element to use. See: https://www.w3schools.com/tags/att_button_form.asp

muratcorlu:
This PR adds `form` property to the button to be able to target an HTML form that is not wrapping the button. I added some documentation as well to explain how to use button with forms as a submitter.

Closes #603

### DIFF
--- a/src/components/button/bl-button.ts
+++ b/src/components/button/bl-button.ts
@@ -90,7 +90,7 @@ export default class BlButton extends LitElement {
    * Sets the type of the button. Set `submit` to use button as the submitter of parent form.
    */
   @property({ type: String })
-  type: 'submit' | null;
+  type: 'submit';
 
   /**
    * Sets button type to dropdown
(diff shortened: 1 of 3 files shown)


src/components/button/bl-button.ts
```diff
@@ -104,6 +104,13 @@ export default class BlButton extends LitElement {
   @property({ type: Boolean, reflect: true })
   autofocus = false;
 
+
+  /**
+   * Sets the associated form of the button. Use when `type` is set to `submit` and button is not inside the target form.
+   */
+  @property({ type: String })
+  form: HTMLFormElement | string;
```
Enes5519, line 112: How to pass an element here? By setting property?

muratcorlu: You can either pass the id of the form element (via attribute or property) or the DOM reference of it (by property). Both would work.

I gave some examples in the storybook docs.

Enes5519 (approved):

Issue #608, opened 2023-05-31
Single Select Clear: Optional select fields should display clear button


ogunb:
### Describe the problem or use case

User should be able to opt out on their choice if the select box is not required.

### Proposed solution

Given required prop is not provided, display clear button.

### Alternatives considered

This could also be implemented as an attribute.

### Additional technical context

We already have the clear button, this shouldn't be a hassle.

muratcorlu:
Looks like a nice suggestion. @buseselvi what do you think to show "clear selection button" to select, even if it's in single selection mode?

buseselvi:
I think it is necessary to add the Clear button to the single select as well. But we need to keep it optional because, for the places like pagination, we bring the select selected and it should never be empty.

muratcorlu:
What about defining a rule like "If it's **not** required, then the clear button should be visible when a selection is done"? If we set a select required with a default selection, then there will be no clear button and it will be not possible to make it invalid (like in pagination).

But I'm not sure about running the same rule on multi-selection mode. 🤔

buseselvi:
I looked at the benchmarks, and the clear button is always visible in multi-select, but it is optional in single-select.

ogunb:
We'll update the if as multiple or not required then.

I'll send a PR for this later today.

PR #610, opened 2023-06-01
feat(select): render remove button on non required selects

ogunb:
This pull request adds remove button for optional single selects.
Fixes: https://github.com/Trendyol/baklava/issues/608


muratcorlu:
We need to add `required` to select in the pagination component. Otherwise, clear button will break it.

muratcorlu:
I'm not sure but maybe we need to consider this as a breaking change. 🤔

ogunb:
Although we expect users to have the required prop, some may not. Our implementation and behavior remain consistent and user-friendly. However, it could be seen as a breaking change given the situation. How should we proceed? Should we introduce a temporary boolean like "clearable" or postpone this development?

muratcorlu:
I think we can take that action. The risk looks very low. If some developers use non-required select with a preselected value and don't want to check if the value is set, the only thing they need to do is to add `required` attribute, which really makes sense.

muratcorlu (approved):

buseselvi (approved):

PR #616, opened 2023-06-03
fix(checkbox): remove gap if no label

Enes5519:
With gap:
image

Without gap:
image

### DIFF
--- a/src/components/checkbox-group/checkbox/bl-checkbox.css
+++ b/src/components/checkbox-group/checkbox/bl-checkbox.css
@@ -20,6 +20,10 @@ label {
   word-break: break-all;
 }
 
+.label[hidden] {
+  display: none;
+}
+
 input {
   appearance: none;
   position: absolute;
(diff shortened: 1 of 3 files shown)


src/components/checkbox-group/checkbox/bl-checkbox.ts
```diff
@@ -147,7 +154,9 @@ export default class BlCheckbox extends FormControlMixin(LitElement) {
           @blur=${this.blur}
         />
         <div class="check-mark">${icon ? html`<bl-icon name="${icon}"></bl-icon>` : null}</div>
-        <span class="label"><slot></slot></span>
+        <span class="label" ?hidden=${!this._hasLabel}
+          ><slot @slotchange=${this.handleSlotChange}></slot
+        ></span>
```
muratcorlu, line 159: I think moving `label` class to slot and removing `span` fixes the issue:

```html
<slot class="label"></slot>
```

No need for internal states and JS codes.

src/components/checkbox-group/checkbox/bl-checkbox.css
```diff
@@ -20,6 +20,10 @@ label {
   word-break: break-all;
 }
 
+.label[hidden] {
+  display: none;
+}
+
```
muratcorlu, line 26: This is also not needed anymore, right?

muratcorlu (approved):

PR #619, opened 2023-06-05
feat(input): add readonly attribute

Enes5519:
I need to this property for build date picker with baklava.

image

### DIFF
--- a/src/components/input/bl-input.ts
+++ b/src/components/input/bl-input.ts
@@ -23,7 +23,7 @@ export default class BlInput extends FormControlMixin(LitElement) {
   static get styles(): CSSResultGroup {
     return [style];
   }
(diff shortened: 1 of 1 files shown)


muratcorlu:
[I was hesitant](https://github.com/Trendyol/baklava/issues/405#issuecomment-1467843925) to add this property without any guidance since I was afraid that this can cause some usability issues and confusion on some use-cases. Either we can allow this first and consider side effects later, or maybe we can have a quick consideration to check if we can add some guide to avoid pitfalls -if any- of using this. @buseselvi

muratcorlu:
Can you add this new parameter to the storybook file as well, so it can be controlled in story canvas in the docs?

muratcorlu (approved):

PR #633, opened 2023-06-15
feat(dialog): add `polyfilled` property for setting native or polyfill

### LINKED ISSUE #617: Toast messages are hiding under dialog backdrop
### Issue description

When we have a toast message to show while the dialog is open, the messages are hidden behind the dialog backdrop.
This is affecting the UX since the toasts like error or success messages are not enough clear to users.

### Baklava Version

v2.2.0-beta.2

Low (slightly annoying)

Enes5519:
A prop has been developed that we can set to use native or polyfill. The fastest solution for managing stacking context in browsers without Popover API. 

Fixes #617

### DIFF
--- a/src/components/dialog/bl-dialog.stories.ts
+++ b/src/components/dialog/bl-dialog.stories.ts
@@ -23,7 +23,7 @@ const meta: Meta = {
     open: {
       control: "boolean",
     },
-    levelled: {
+    polyfilled: {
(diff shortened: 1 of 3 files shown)


muratcorlu (approved):
I think we are good to go with this for now. When we solve this issue within our popover component (with Popover API), we may deprecate this property in upcoming major releases.

PR #355, opened 2023-01-05
fix(select): visual fixes over long texts

muratcorlu:
This fixes several visual issues about checkbox and select components that is reported in #353 and #352 

* Checkbox checkmark background color
* Checkbox checkmark deformation with limited space
* Checkbox label with long text and multiple lines
* Select component long placeholder and labels

I tried to add extra stories to our storybook to cover those situations in our visual tests.

Fixes #353 
Fixes #352

PR #636, opened 2023-06-16
fix(radio): radio button flex-shrink problem

doganozturk:
I've noticed an issue with the `bl-radio` component where the radio button's size shrinks if the text within the `#label` element becomes too long. The `#label` element is a flex container, its `:before` pseudo-selector, which is used to create the radio button, is designed with strict width and height settings, leading to this behavior.

The cause of the radio button shrinking is tied to its `flex-shrink` property. However, applying `flex-shrink: 0;` to address this issue would be overly specific.

As a solution, I've implemented a `flex: none;` setting which prevents the radio button from shrinking.

I'm attaching a few example screenshots to illustrate the issue:

246438482-cfee32d8-0a4d-4ddc-83c0-a3277cb34112

246438494-06740896-5d8e-44d7-a166-7ea6b3c36d32

### DIFF
--- a/src/components/radio-group/radio/bl-radio.css
+++ b/src/components/radio-group/radio/bl-radio.css
@@ -24,8 +24,10 @@
   content: '';
   display: inline-block;
   box-sizing: border-box;
-  width: var(--size);
-  height: var(--size);
+  min-width: var(--size);
+  min-height: var(--size);
+  max-width: var(--size);
+  max-height: var(--size);
   background-color: white;
   border-radius: var(--bl-border-radius-circle);
   border: solid 1px var(--bl-color-neutral-lighter);


muratcorlu:
We had [similar issue with checkbox component ](https://github.com/Trendyol/baklava/issues/352) and fixed it by [setting min/max width/height](https://github.com/Trendyol/baklava/pull/355/files) for it. Can we have some side effects of disabling flex layout for radio in terms of vertical alignment or gap between label and it etc?

doganozturk:
> Thanks for the PR and very good explanation! 😊
> 
> We had [similar issue with checkbox component ](https://github.com/Trendyol/baklava/issues/352) and fixed it by [setting min/max width/height](https://github.com/Trendyol/baklava/pull/355/files) for it. Can we have some side effects of disabling flex layout for radio in terms of vertical alignment or gap between label and it etc?

Same effect seems to be achieved via `min/max` `width/height` as it seems. I can update my PR according to that solution. Would it be more suitable?

```css
#label::before {
  content: '';
  display: inline-block;
  box-sizing: border-box;
  min-width: var(--size);
  min-height: var(--size);
  max-width: var(--size);
  max-height: var(--size);
  ...
}
```

muratcorlu:
I'm not completely sure as well but even if there is no difference, to be aligned with other components, that would be better, I believe. Thanks!

muratcorlu (approved):

PR #637, opened 2023-06-17
fix(tooltip): remove extra space

Enes5519:
Screen Shot 2023-06-17 at 16 05 39

the top input is the expected state and the bottom is the actual state. (look right gap)

### DIFF
--- a/src/components/tooltip/bl-tooltip.css
+++ b/src/components/tooltip/bl-tooltip.css
@@ -7,11 +7,6 @@
   cursor: pointer;
 }
 
-.wrapper{
-  display: inline-block;
-  color: var(--bl-color-neutral-full);;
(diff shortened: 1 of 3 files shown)


src/components/tooltip/bl-tooltip.css
```diff
@@ -7,11 +7,6 @@
   cursor: pointer;
 }
 
-.wrapper{
-  display: inline-block;
-  color: var(--bl-color-neutral-full);;
```
muratcorlu, line 12: Don't we need this color definition? 🤔

muratcorlu: Ah, we already have it on `.content` selector. :+1:

PR #640, opened 2023-06-19
fix(radio): fix height limit and block flow improvements

### LINKED ISSUE #597: Radio group and radio styling issues
In row layout, flexbox does not spread for the remaining space, this can create problems in wider forms. We partially solved this issue by giving bl-radio-group display: block style; however in order to sprea bl-radio does not share same width with bl-radio group element thus, it stays with its own content width. We could only widen it with min-width but I don't think this is actually a valid solution or best practice.

Also since radio elements, label element has height has been set, is creates visual deficiencies when having a label that has more than 16px height.

stovein:
Fixes #597


muratcorlu:
I fixed the styling issues about vertical alignment. I added a new CSS variable to be able to set vertical alignment of the radio button. By default, I set it as vertically centered. @buseselvi Please check if those are good for you.

buseselvi (approved):

omrumbakitemiz:
👀 👀

muratcorlu:
Do we have any concerns about this PR @stovein ? Can we merge this?

stovein:
We have a slight radio group alignment problem, since our radio groups host is now in flow layout but options classed div is in flex layout, we cannot directly manipulate alignment from host via giving classes or styles. We could add some measures for that and I think it will be ok and configurable most of the time.

An example of misalignment
image

buseselvi:
Title and radio inputs are looking a bit separated as an example. Can we show an example with 2-3 lines of text aligned middle and side by side, which looks more aesthetically pleasing? Or maybe just the explanation without an example in the storybook 🤔 

image

stovein:
How is new sotry, if its ok I believe having a story would be better to get a feel how it works.

erbilnas (approved):

buseselvi:
It looks perfect 🙏🏻 But now it is not working when clicking on that radio input in the storybook, i couldn't select them. @stovein

stovein:
since problem in the link provided is not related to this pr, we will continue from #659 .

Issue #627, opened 2023-06-12
[Feature Request]: Padding or remove alert icon for input


Enes5519:
### Describe the problem or use case

I want remove icon of invalid input or change paddings of input.

![resim](https://github.com/Trendyol/baklava/assets/10272532/92b77bd6-92fa-42a3-8503-8ae1d51f81a7)

AykutSarac:
Closing this in favor of #643 and #642.

PR #642, opened 2023-06-22
feat(input): add attribute for remove icon

Enes5519:
I want to add my custom icon with position: absolute. So I need this attribute.
resim

### DIFF
--- a/src/components/input/bl-input.ts
+++ b/src/components/input/bl-input.ts
@@ -127,6 +127,12 @@ export default class BlInput extends FormControlMixin(LitElement) {
   icon?: string;
 
   /**
+   * Forcefully removes the icon.
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'remove-icon' })
+  removeIcon = false;
+
+  /**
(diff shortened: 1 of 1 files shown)


Enes5519:
Fixes #627

muratcorlu:
Since it seems you need this functionality just for repositioning the error icon, do you still need this, when we have #643?

Enes5519:
I want to put an icon with a tooltip. That's why I don't want it to be mixed with the normal icon.

erbilnas (changes requested):
@ogunb's last commit  #649 affects here, so this needs to be refactored accordingly

ogunb:
After merging #649 would we still need this attribute? 

I believe, we can solve the problem here with slot icon and some margin. And If we want to remove the default icon anyway we can just provide an empty slot.

muratcorlu:
@Enes5519 can we close with in favor of #649 ?

muratcorlu:
Closing this. Please reopen if you still find this relevant. @Enes5519

PR #643, opened 2023-06-22
feat(input): set padding-horizontal values with css properties

Enes5519:
resim

For #627

### DIFF
--- a/src/components/input/bl-input.ts
+++ b/src/components/input/bl-input.ts
@@ -18,8 +18,8 @@ export type InputSize = 'small' | 'medium' | 'large';
  * @tag bl-input
  * @summary Baklava Input component
  *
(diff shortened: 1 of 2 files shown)


src/components/input/bl-input.ts
```diff
@@ -17,6 +17,9 @@ export type InputSize = 'small' | 'medium' | 'large';
 /**
  * @tag bl-input
  * @summary Baklava Input component
+ *
+ * @cssproperty [--bl-input-padding-left] Sets the padding left
+ * @cssproperty [--bl-input-padding-right] Sets the padding right
```
muratcorlu, line 22: We'll need to start considering right-to-left languages soon, so let's start using logical property naming here like `--bl-input-padding-start` and `--bl-input-padding-end`. You can update the description accordingly.

muratcorlu (approved):

PR #649, opened 2023-07-11
Implement Icon slot for bl-input

### LINKED ISSUE #559: Input should accept Icon slot
As a next step of #558 and #525, it would be obvious to expect the same capabilities for our Input component. Currently we only able to set `name` of the icon for input.

```html
<bl-input>
	<bl-icon slot="icon"><svg ... /></bl-icon>
</bl-input>
```

Our button component already accepts both `icon` attribute and an `icon` slot. So I think we can keep this approach also for input.

**Additional context**
I think this feature would be used for the real requirements and we should not encourage people to use slots over attributes for simple/default icons.

ogunb:
Closes #559

### DIFF
--- a/src/components/input/bl-input.stories.mdx
+++ b/src/components/input/bl-input.stories.mdx
@@ -86,27 +86,6 @@ import { extraPadding } from '../../utilities/chromatic-decorators';
 export const BlButton = args => html`
   <bl-button>Check</bl-button>`
 
-export const SpinnerIcon = args => html`
-  <style>
-    .spinner {
-      animation: spin 1s linear infinite;
-      color: var(--bl-color-primary);
-    }
-
-    @keyframes spin {
-      0% {
-        transform: rotate(0deg);
-      }
-
-      100% {
-        transform: rotate(360deg);
-      }
-    }
-  </style>
-
-  <bl-icon slot="icon" name="loading" class="spinner"></bl-icon>
-`
-
 export const SingleInputTemplate = (args) => html`<bl-input
     type='${ifDefined(args.type)}'
     label='${ifDefined(args.label)}'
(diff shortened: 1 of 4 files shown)


src/components/input/bl-input.stories.mdx
```diff
@@ -86,6 +86,27 @@ import { extraPadding } from '../../utilities/chromatic-decorators';
...
+    @keyframes spin {
+      0% {
+        transform: rotate(0deg);
+      }
+
+      100% {
+        transform: rotate(360deg);
+      }
+    }
+  </style>
```
muratcorlu, line 105: Putting style tag inside `bl-input` would misguide the people. I think it's better to put style tag outside.

src/components/input/bl-input.ts
```diff
@@ -33,6 +33,9 @@ export default class BlInput extends FormControlMixin(LitElement) {
   @query('input')
   validationTarget: HTMLInputElement;
 
+  @query('slot[name="icon"]')
+  iconSlot: HTMLSlotElement;
```
muratcorlu, line 37: This can be a private getter instead of public property, like we did in Button component.

```ts
  private get _hasIconSlot() {
    return this.querySelector(':scope > [slot="icon"]') !== null;
  }
```

muratcorlu (approved):

muratcorlu:
@buseselvi There are 2 new stories for the input component. Can you check if they are ok for you?

AykutSarac (approved):

erbilnas (approved):

buseselvi:
Can we show the spinner color by the input states in the storybook? Of course we can make different colors, but I wanted to ask to make it look consistent in our document.
image

ogunb:
> Can we show the spinner color by the input states in the storybook? Of course we can make different colors, but I wanted to ask to make it look consistent in our document. image

Sure, I removed the color attribute.

buseselvi (approved):

erbilnas (approved):

PR #650, opened 2023-07-12
feat(react): export react component types

### LINKED ISSUE #307: Export component types
Sometimes when creating a custom component using existing Baklava component you could spread props. It would be great to import types of e.g "BlInputType" so we could use it as a type and spread. I'm open for suggestions if any :)

Export `BlInputType` and other component types from `baklava-react.d.ts`

AykutSarac:
This PR adds support to use React components both as value and type.

Example usage:
```tsx
import { BlDialog } from "@trendyol/baklava/dist/baklava-react";

const Component = () => {
  const dialogRef = React.useRef<BlDialog>(null);

  return (
    <BlDialog ref={dialogRef}>...</BlDialog>
  )
}
```

Fixes #307

### DIFF
--- a/scripts/generate-react-exports.js
+++ b/scripts/generate-react-exports.js
@@ -15,11 +15,14 @@ const importStatements = [
   'import { ISelectOption } from "./components/select/bl-select"',
 ];
 
+const exportStatements = [];
+
 function writeBaklavaReactFile(fileContentParts) {
   let fileContentText =
     `/* eslint-disable @typescript-eslint/ban-ts-comment */\n` +
     `// @ts-nocheck\n` +
     `${importStatements.join('\n')}\n\n` +
+    `${exportStatements.join('\n')}\n\n` +
     `${fileContentParts.join('\n\n')}\n`;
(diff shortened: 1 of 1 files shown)


muratcorlu:
Do you see any risk of this change to break current React implementations?

AykutSarac:
> Do you see any risk of this change to break current React implementations?

No, it only brings type support which was previously inferred as `any`. There shouldn't be any breaking change.

erbilnas (approved):

muratcorlu (approved):

scripts/generate-react-exports.js
```diff
@@ -77,6 +80,7 @@ for (const module of customElementsModules) {
   const Type = componentName + 'Type';
 
   importStatements.push(`import type ${Type} from "./${importPath}";`);
+  exportStatements.push(`export declare type ${Type.replace('Type', '')} = ${Type}`);
```
Enes5519, line 83: You can use `componentName` variable instead of `${Type.replace('Type', '')}`.

AykutSarac: Nice catch, updated.

PR #651, opened 2023-07-12
fix(input): remove type from input min/max

### LINKED ISSUE #612: [Feature Request]: Date Input min/max Props

Currently there is no date input in baklava. But when we give date as input type, we can use native date as input. However, at this point, we cannot give the min and max props that we give to the native date inputs. Because when we give, the value given is cast to number and the value is converted to NaN.

### Proposed solution

Making the min and max props also accept the `YYYY-MM-DD` pattern.

### Alternatives considered

_No response_

_No response_

AykutSarac:
This PR removes @property type from min and max attributes, thus it will allow input type="date" to be parsed correctly.

Fixes #612

### DIFF
--- a/src/components/input/bl-input.ts
+++ b/src/components/input/bl-input.ts
@@ -84,13 +84,13 @@ export default class BlInput extends FormControlMixin(LitElement) {
   /**
    * Sets the smallest number can be entered to a `number` input
    */
-  @property({ type: Number, reflect: true })
+  @property({ reflect: true })
   min?: number;
 
   /**
    * Sets the biggest number can be entered to a `number` input
    */
-  @property({ type: Number, reflect: true })
+  @property({ reflect: true })
   max?: number;
 
   /**


mertcanaltin:
good luck to your hands 🎸 

 it is important to consider the expected type of the property and whether other components or code segments that use this property will be affected by this change. If the property doesn't already represent a numeric value or if there are no type dependencies in other parts, this change generally shouldn't create a problem

src/components/input/bl-input.ts
```diff
@@ -84,13 +84,13 @@ export default class BlInput extends FormControlMixin(LitElement) {
   /**
    * Sets the smallest number can be entered to a `number` input
    */
-  @property({ type: Number, reflect: true })
+  @property({ reflect: true })
   min?: number;
```
muratcorlu, line 88: Don't we need to set a union type as `number | string` to allow date style input for this?

AykutSarac: I've additionally added 'date' to type now.

erbilnas (approved):

muratcorlu (approved):

Enes5519 (approved):

PR #663, opened 2023-07-18
feat(select): add select clearable prop

AykutSarac:
This PR fixes the breaking change:
- Select component shouldn't have clear button by default, instead `clearable` attribute should be given to set it clearable

### DIFF
--- a/src/components/select/bl-select.stories.mdx
+++ b/src/components/select/bl-select.stories.mdx
@@ -157,10 +157,7 @@ The select component includes a clear button. Clear button can be displayed by p
   <Story name="Select With Clear Button" args={{ placeholder: "Choose country", value: 'tr', clearable: true }} play={selectOpener}>
     {SelectTemplate.bind({})}
   </Story>
-  <Story name="Required Select Without Clear Button" args={{ placeholder: "Choose country", value: 'tr', clearable: true }} play={selectOpener}>
-    {SelectTemplate.bind({})}
-  </Story>
-  <Story name="Required Multiple Select With Clear Button" args={{ placeholder: "Choose countries", value: ['nl'], multiple: true, clearable: true  }} play={selectOpener}>
+  <Story name="Multiple Select With Clear Button" args={{ placeholder: "Choose countries", value: ['nl'], multiple: true, clearable: true  }} play={selectOpener}>
     {SelectTemplate.bind({})}
   </Story>
 </Canvas>
--- a/src/components/select/bl-select.stories.mdx
+++ b/src/components/select/bl-select.stories.mdx
@@ -34,6 +34,9 @@ import { centeredLayout } from '../../utilities/chromatic-decorators';
     required: {
       control: 'boolean'
     },
+    clearable: {
+      control: 'boolean'
+    },
     disabled: {
       control: 'boolean'
     },
@@ -90,6 +93,7 @@ export const SelectTemplate = (args) => html`<bl-select
     label=${ifDefined(args.label)}
     ?label-fixed=${args.labelFixed}
     ?multiple=${args.multiple}
+    ?clearable=${args.clearable}
     ?required=${args.required}
     ?disabled=${args.disabled}
     ?success=${args.success}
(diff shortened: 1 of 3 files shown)


fatihhayri (approved):

erbilnas (approved):

src/components/select/bl-select.stories.mdx
```diff
@@ -147,17 +151,16 @@ Selected options will be visible on input seperated by commas.
...
-However, in multiple select, the clear button is always shown, irrespective of the `required` attribute.
+The select component includes a clear button. Clear button can be displayed by passing `clearable` attribute to the Select component.
 
 <Canvas>
-  <Story name="Select With Clear Button" args={{ placeholder: "Choose country", value: 'tr' }} play={selectOpener}>
+  <Story name="Select With Clear Button" args={{ placeholder: "Choose country", value: 'tr', clearable: true }} play={selectOpener}>
     {SelectTemplate.bind({})}
   </Story>
-  <Story name="Required Select Without Clear Button" args={{ placeholder: "Choose country", value: 'tr', required: true }} play={selectOpener}>
+  <Story name="Required Select Without Clear Button" args={{ placeholder: "Choose country", value: 'tr', clearable: true }} play={selectOpener}>
```
ogunb, line 160: Don't need this anymore.

src/components/select/bl-select.test.ts
```diff
@@ -136,8 +128,18 @@ describe('bl-select', () => {
...
+    const el = await fixture<BlSelect>(html`
+      <bl-select clearable>
+        <bl-select-option value="1">Option 1</bl-select-option>
+        <bl-select-option value="2" selected>Option 2</bl-select-option>
+      </bl-select>
+    `);
+    const removeAll = el.shadowRoot?.querySelector<BlButton>('.remove-all');
+    expect(removeAll).to.exist;
+  });
   it('should remove selected option on remove all click on non-required single select', async () => {
```
ogunb, line 141:
```suggestion
  it('should remove selected option on remove all click on single select with clearable prop', async () => {
```

src/components/select/bl-select.ts
```diff
@@ -292,14 +291,16 @@ export default class BlSelect<ValueType extends FormValue = string> extends Form
...
     </ul>`;
-    const removeButton = html`<bl-button
-      class="remove-all"
-      size="small"
-      variant="tertiary"
-      kind="neutral"
-      icon="close"
-      @click=${this._onClickRemove}
-    ></bl-button>`;
+    const removeButton = this.clearable
```
ogunb, line 294: Multiple selects should always be clearable.

src/components/select/bl-select.test.ts
```diff
@@ -154,7 +156,7 @@ describe('bl-select', () => {
     expect(el.value).to.null;
   });
   it('should remove selected options', async () => {
-    const el = await fixture<BlSelect>(html`<bl-select multiple>
+    const el = await fixture<BlSelect>(html`<bl-select clearable multiple>
```
ogunb, line 159: Shouldn't need this.

ogunb (changes requested):
We are missing clear button on multiple selects with these changes.
https://6267ebb4dd54f5004a65f14a-ncjumrcdek.chromatic.com/?path=/docs/components-select-select--documentation#multiple-select

I don't know why chromatic doesn't show that.

--- 

Also I think we should discuss whether a design system should decide when to render clear button or to have a prop for it.

Previous discussions:
https://github.com/Trendyol/baklava/pull/610
https://github.com/Trendyol/baklava/issues/608

src/components/select/bl-select.ts
AykutSarac: Why is that?

AykutSarac: There could be cases like an option is selected by default and disabled, thus cannot be de-selected. If we display clear button it will clear that too. `clearable` prop should be fully optional.

ogunb: Apparently it's because of benchmarks: https://github.com/Trendyol/baklava/issues/608#issuecomment-1571508124

Also it's introducing another breaking change while fixing one. It should at least default to true.

ogunb: Oh, I though you meant the input, sorry. 

I don't quite see how this case would solve it. Since the data flow is unidirectional, developer could set required items back on clear. (Assuming there aren't any bugs with updating the value.)

It comes to my initial question actually. Should the design team decide when to render the clear button or the user?

If we area going with clearable prop (which I think is superior option for DX but not sure on design aspect of things.) it's default status should act like before:
- True by default on multiple select. User can provide false if needed.
- False by default on single select. User can provide true if needed.

ogunb: Hey @buseselvi, do you have any comments on this?

buseselvi: I think we should keep multiple select as clearable true since there is no breaking change and it is more accurate in terms of UX.

ogunb (approved):

buseselvi (approved):

muratcorlu:
Sorry to being late to the discussion but I think this change proposes bad UX in the case of "non-required and non-clearable" selects. User initially don't need to make a selection but once they make, then there is no returning back.

I think instead we should promote using "required" attribute properly [as we already discussed in the PR we added this feature](#610).

AykutSarac:
What do I do if I want to display clear button to user but select is also required? I believe the prop was necessary.

muratcorlu:
In single selection mode with a required select, what is the purpose of clearing the selection? You already need to make selection anyway?

AykutSarac:
Considering the complexity and diversity of real-world business flows, it is challenging to anticipate every scenario that users might encounter.

Just an example that came into my mind:
Imagine a real-time based scenario where the user may select an option (optional), however after the selection I don't want the option to be cleared. There are many cases that you cannot cover, so this should solely depend on the developer.
Or there could be an input depends on the select component, if it has value it will take specific action and don't want it to be cleared back, whilst it could be optional at beginning.

Providing a "clearable" prop gives developers the flexibility to customize the behavior of the select component based on specific use cases. Some situations might require a clear button, while others may not. Giving control to the developers allows them to align the component with the project's requirements.

Issue #670, opened 2023-07-26
[BUG]: Unable to select the radio option at the bottom when scrolling dialog through the bottom


MehmetCanBOZ:
### Issue description

Example code sandbox : https://codesandbox.io/s/competent-cori-v2y43p

1- open example code sandbox
2- click the "Open Dialog" button
3- scroll the dialog through the bottom
4- try to select "option 25"
5- Unable to select the radio option, it scrolls up.

### Baklava Version

2.2.0

### Operating system

- OS: [e.g. iOS]:    iOS
- Browser [e.g. chrome, safari]: chrome

### Priority this issue should have

Medium (should be fixed soon)

### Please review the checkboxes that are applicable.

- [ ] This issue is directly related to React

MehmetCanBOZ:
Sorry for the confusion, I think this bug is about the **bl-radio-group**, not dialog.

fatihhayri:
This is a very quirk bug. :(

MehmetCanBOZ:
I think that it is about the focusedOptionIndex in BlRadioGroup. When the user scrolls and click the option at the bottom, the focusedOptionIndex does not change and it is initially 0. So when the user clicks the option, handleFocus function calls and it focuses on the first option. So it scrolls up.

PR #672, opened 2023-07-27
fix(radio): focused option element on BlRadioGroup

MehmetCanBOZ:
This pull request fixes the focused option element on BlRadioGroup
Fixes: https://github.com/Trendyol/baklava/issues/670

### DIFF
--- a/src/components/radio-group/bl-radio-group.ts
+++ b/src/components/radio-group/bl-radio-group.ts
@@ -100,12 +100,25 @@ export default class BlRadioGroup extends FormControlMixin(LitElement) {
     event.preventDefault();
   }
 
+  getFocusedOptionIndex() {
+    return this.focusedOptionIndex;
+  }
+
   connectedCallback(): void {
     super.connectedCallback();
 
     this.tabIndex = 0;
     this.addEventListener('focus', this.handleFocus);
     this.addEventListener('keydown', this.handleKeyDown);
+    this.addEventListener('mouseover', this.handleMouseOver);
+  }
+
+  private handleMouseOver(event: MouseEvent) {
+    const mouseOverElement = event.target;
+
+    if (mouseOverElement instanceof BlRadio) {
+      this.focusedOptionIndex = this.availableOptions.indexOf(mouseOverElement);
+    }
   }
 
   private handleFocus() {
--- a/src/components/radio-group/bl-radio-group.ts
+++ b/src/components/radio-group/bl-radio-group.ts
@@ -65,6 +65,8 @@ export default class BlRadioGroup extends FormControlMixin(LitElement) {
 
   private focusedOptionIndex = 0;
 
+  private isMouseEventFocus = false;
+
   private handleOptionChecked(event: CustomEvent) {
     const checkedOption = event.target as BlRadio;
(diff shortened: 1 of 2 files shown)


fatihhayri (approved):

ogunb (approved):

erbilnas (approved):

src/components/radio-group/bl-radio-group.ts
```diff
@@ -100,12 +100,25 @@ export default class BlRadioGroup extends FormControlMixin(LitElement) {
...
     this.addEventListener('focus', this.handleFocus);
     this.addEventListener('keydown', this.handleKeyDown);
+    this.addEventListener('mouseover', this.handleMouseOver);
+  }
+
+  private handleMouseOver(event: MouseEvent) {
+    const mouseOverElement = event.target;
+
+    if (mouseOverElement instanceof BlRadio) {
+      this.focusedOptionIndex = this.availableOptions.indexOf(mouseOverElement);
```
muratcorlu, line 120: I don't think this is a good idea. Because now we set "focused option index" without actually focusing but just hovering it.

src/components/radio-group/bl-radio-group.ts
```diff
@@ -100,12 +100,25 @@ export default class BlRadioGroup extends FormControlMixin(LitElement) {
...
+    return this.focusedOptionIndex;
+  }
+
   connectedCallback(): void {
     super.connectedCallback();
 
     this.tabIndex = 0;
     this.addEventListener('focus', this.handleFocus);
     this.addEventListener('keydown', this.handleKeyDown);
+    this.addEventListener('mouseover', this.handleMouseOver);
```
muratcorlu, line 113: Mouseover event will trigger with every pixel move over a bl-radio-group element, which seems very inefficient. I think, we should question why do we need that focused option index at all instead.

muratcorlu (changes requested):
I don't think proposed solution is a good one and actually solves the problem. I suggest to do some more investigation.

MehmetCanBOZ:
> I don't think proposed solution is a good one and actually solves the problem. I suggest to do some more investigation.

Thanks for your reviews, they are so helpful. I thought that it is about the **focusedOptionIndex**. Because when the user scrolls and click the option at the bottom, the **focusedOptionIndex** does not change and it is initially 0. So when the user clicks the option, the handleFocus function calls and it focuses on the first option. So the user can not select the option. I thought we need the change **focusedOptionIndex** for this problem. My solution solves the problem but as you said and I agree, it is not a proper solution for this. Another solution can be implemented.

src/components/radio-group/bl-radio-group.ts
MehmetCanBOZ: The focusedOptionIndex is for keyboard navigation, but if the user clicks the option with the mouse **handleFocus** is calling so it focuses on the first option. Do we need **handleFocus** in here?

MehmetCanBOZ: I used a flag named **isMouseEventFocus** to track whether the focus is due to a mouse event or not. So handleFocus is executed only with keyboard navigation.

PR #677, opened 2023-08-01
fix(select): return single value for single select

### LINKED ISSUE #675: [BUG]: bl-select should return value instead of array on single select
### Issue description

BlSelect currently returns array on `e.detail` even if it's single select which doesn't make sense. It should only return array on multiselect.

Additionally, this broke the component when **version 2.2.0** was introduced with **clear button** as my initial value was empty string instead of null, and it was getting `e.detail[0].value`.

### Reproduction

```html
<bl-select bl-select="print" name="country" label="Country" help-text="Please select the country you live in" required>
  <bl-select-option value="tr">Turkiye</bl-select-option>
  <bl-select-option value="nl">The Netherlands</bl-select-option>
</bl-select>

<script>
  // print's bl-select e.detail value
  document.querySelector("bl-select").addEventListener('bl-select', (e) => {
    console.log(e.detail.value); // output: undefined 
    console.log(e.detail); // output: [{...}]
  });
</script>
```

### Media & Screenshots

https://github.com/Trendyol/baklava/assets/47941171/28617b50-5859-4c0d-9a15-c0b58a876a29

### Baklava Version

2.3.0-beta.2

### Operating system

- OS: [e.g. iOS]: MacOS
- Browser [e.g. chrome, safari]: Brave

Low (slightly annoying)

AykutSarac:
fixes #675


muratcorlu:
Do you think we should make a major release (v3) for this?

AykutSarac:
I believe we should at least have one new component per major release or significant change, until that time such changes can wait unless it requires urgent fix.

muratcorlu:
Stable releases should last longer than a few months. Our consumers should not need to make a careful upgrade every few months. So I propose avoiding breaking changes as much as possible for a while, and collect those proposals in a backlog for a bigger major release.

We may consider to have another branch for major releases like `next-major` and we can merge breaking changes to that branch, but this will also add extra maintenance cost since we will need to sync other works from `next` branch to `next-major` continuously.

muratcorlu:
Yes, `beta` releases can have major releases. But we currently have only a single beta channel. If we merge this branch with a "breaking change" notice in commit message, then our next branch will release "v3.0.0-beta.1". Then when we merge next to main, we'll have a major stable release. there will be no chance to make a "v2.3.0" release.

leventozen (approved):

PR #689, opened 2023-09-07
fix(tab): changing the tooltip button in the tab with help text to an…

serhatakti:
image

The hover background of the button used for the tooltip is sticking to the text. There is no click action on the button. That's why I think it's unnecessary

### DIFF
--- a/src/components/tab-group/tab/bl-tab.ts
+++ b/src/components/tab-group/tab/bl-tab.ts
@@ -112,13 +112,7 @@ export default class BlTab extends LitElement {
     const helpTooltip = this.helpText
       ? html` <div class="help-container">
           <bl-tooltip>
-            <bl-button
-              slot="tooltip-trigger"
-              icon="info"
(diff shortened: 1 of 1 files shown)


muratcorlu:
Hi @serhatakti. Thanks for the proposal. Since "hover" is also a pointer interaction (and it's triggered by tap on touch screen devices) we wanted to keep clickable area bigger then icon by using button. Without a button, we'll need to handle a11y and UX issues for a clickable (or hoverable) icon. Maybe @buseselvi also wants to add her comments.

buseselvi:
Hi @serhatakti , as @muratcorlu said this is the hover state of the button not pressed so that the tooltip is working while hovering, it is necessary to show the hover state to inform users that there is a button that can be interacted with. But also I think you are right about that there should be some padding between the tab title and the button, we can give that. 🚀

leventozen:
Since, there is no further improvement on this one, I am closing this PR.

PR #692, opened 2023-09-08
feat(pagination): use auto width on select

### LINKED ISSUE #690: bl-pagination: can we change the bl-select size based on a variable

our bl-select content larger than 128px so we need to extend it

### Proposed solution

_No response_

### Alternatives considered

_No response_

_No response_

IbrahimCagriKoca:
With the help of #706, this PR sets width of the pager select in pagination component to `auto`. That helps to see label of current selection in case of custom texts.

Closes #690


muratcorlu:
I proposed a PR to add a feature to our select component to have automatic width. #706 If that PR is good to go, I would propose to revert this variable (since I can not see any other use case) and just set select width to auto in the pagination component.

buseselvi (approved):

muratcorlu (approved):

PR #696, opened 2023-09-08
chore(deps): lit version upgrade

muratcorlu:
This PR upgrades Lit to 2.8

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "@open-wc/form-control": "^0.7.0",
         "@open-wc/form-helpers": "^0.2.2",
         "element-internals-polyfill": "^1.2.3",
-        "lit": "^2.6.1"
+        "lit": "^2.8.0"
       },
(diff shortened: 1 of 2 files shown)


AykutSarac:
I think we can upgrade `@lit-labs/react` with this PR.

muratcorlu:
@AykutSarac there is a breaking change on `@lit-labs/react` library which needs some code changes. I think it'll be better to have it in a separate PR. https://github.com/lit/lit/blob/main/packages/labs/react/CHANGELOG.md

AykutSarac (approved):

AykutSarac:
> @AykutSarac there is a breaking change on `@lit-labs/react` library which needs some code changes. I think it'll be better to have it in a separate PR. https://github.com/lit/lit/blob/main/packages/labs/react/CHANGELOG.md

IIRC I've tried with the latest including the breaking change and it didn't effect the output. I agree it better be in a separate PR.

PR #698, opened 2023-09-11
 fix(docs): fix broken link to Contribution Guideline document

onurravli:
In GitHub read me file the Contribution Guideline Document's link was working, because they were in the same directory. But in Storybook, it wasn't working. So I've made an update for this link.


muratcorlu:
I would be great if we can use [Storybook page](https://baklava.design/?path=/docs/documentation-contributing-baklava-contribution-guideline--documentation) in storybook link and GitHub page in GitHub. I'm not sure if we can easily achieve this. Otherwise we link both of them to a full url, of course.

onurravli:
What about using 2 README files as a workaround? Like [this](https://github.com/onurravli/baklava/commit/6c95fff9c0e2dbd393b9f72406fab2c2df718d66) commit.

PR #705, opened 2023-09-14
feat(popover): allow nested popovers

### LINKED ISSUE #703: [Feature Request]: Allow nested popovers

Current implementation allows only one popover to be active on the page at a time. That means you cannot use nested popovers and it prevents you from using tooltip inside a popover or create nested menus.

https://codesandbox.io/s/sharp-cookies-rw76y8?file=/index.html

### Proposed solution

_No response_

### Alternatives considered

_No response_

_No response_

ogunb:
This PR will allow users to nest popovers inside of each other. Current implementation doesn't allow the use of `bl-tooltip` inside a popover since `bl-tooltip` also uses popover component. 

Closes #703

### DIFF
--- a/src/components/popover/bl-popover.stories.mdx
+++ b/src/components/popover/bl-popover.stories.mdx
@@ -37,6 +37,11 @@ export const popoverOpener = async ({canvasElement}) => {
   }
 }
 
+export const nestedPopoverOpener = async ({canvasElement}) => {
+  await popoverOpener({ canvasElement });
+  await popoverOpener({ canvasElement: canvasElement.querySelector('#popover-nested-parent') });
+}
+
 export const handleClick = async (event,args) => {
   const popover = document.getElementById(`popover-${ifDefined(args.id)}`);
   popover.target = document.getElementById(`button-${ifDefined(args.id)}`);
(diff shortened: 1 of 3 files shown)


erbilnas (approved):

src/components/popover/bl-popover.ts
```diff
@@ -219,7 +219,12 @@ export default class BlPopover extends LitElement {
 
   private _handlePopoverShowEvent(event: Event) {
     if (event.target !== this) {
-      this.hide();
+      const { parentElement } = event.target as HTMLElement;
+      const hasPopoverParent = parentElement?.closest("bl-popover") === this;
```
muratcorlu, line 223: Does this work with a 3rd level nested popover? Maybe we could consider something like:

```ts
const isNestedPopover = this.contains(parentElement);
```

src/components/popover/bl-popover.stories.mdx
```diff
@@ -37,6 +37,11 @@ export const popoverOpener = async ({canvasElement}) => {
   }
 }
 
+export const nestedPopoverOpener = async ({canvasElement}) => {
+  await popoverOpener({ canvasElement });
+  await popoverOpener({ canvasElement: canvasElement.querySelector('#popover-nested-parent') });
```
muratcorlu, line 42: I think we can add a small delay before opening this second popover since it causes positioning failure on chromatic firefox screenshot. Maybe 100ms... Something like:

```ts
await new Promise((resolve) => setTimeout(resolve, 100));
```

src/components/popover/bl-popover.ts
ogunb: Totally missed that. That was a null check before. Thanks.

muratcorlu (approved):
Good job! 🚀

buseselvi (approved):

PR #706, opened 2023-09-15
feat(select): work with auto width

### LINKED ISSUE #693: Select auto-width
Select component optionally should able to work with `width: auto`. In this case width of the select component should be as small as fitting the selected option's label.

In this case select should also respect `max-width` definition to keep itself usable in case it has very long option labels.

muratcorlu:
This PR adds the capability to our select working with `auto` sizing. If you set `width: auto;` to select component, it's width becomes fluid respected to its current label, placeholder, or selected value.

Closes #693


buseselvi (approved):

AykutSarac (approved):

AykutSarac:
I noticed there seem to be some overflow problem with the sizing. Do you think it should have the size of max width of the select option or the select itself?

image

muratcorlu:
If options are wider than select, popover should be bigger, I think that was our decision and we applied this in dropdown. Maybe we need a fix for select as well. Do you agree @buseselvi ?

buseselvi:
Yes, it would be nice to give the popover the ability to be wider than the input, like in the dropdown button. I think we will also need this when developing the date picker component in the future.

PR #707, opened 2023-09-15
fix(input): revalidate input on prop changes

### LINKED ISSUE #704: [BUG]: Input validation should be rechecked on attribute changes
### Issue description

I set a value of the input something invalid, I remove the validations and set to disabled. The error message is still shown. It should be rechecked on every attribute change.

### Media & Screenshots

image

### Baklava Version

2.3.0-beta.3

### Operating system

_No response_

Low (slightly annoying)

muratcorlu:
This PR updates input with every attr/prop changes. So validation can be retriggered in case of validation rule changes, without updating the value.

Fixes #704

### DIFF
--- a/src/components/input/bl-input.ts
+++ b/src/components/input/bl-input.ts
@@ -285,9 +285,9 @@ export default class BlInput extends FormControlMixin(LitElement) {
       this.setValue(this.value);
 
       await this.validationComplete;
-
-      this.requestUpdate();
     }
+
+    this.requestUpdate();
   }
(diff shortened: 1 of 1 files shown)


Enes5519:
I have limited information about Framework, so I have to ask questions.
Can't we handle this in Litjs without requestUpdate?
If we don't normally use the `updated` function, does the Lit automatically trigger the requestUpdate function?

AykutSarac (approved):

PR #709, opened 2023-09-19
chore(react): add jsdoc & upgrade lit

AykutSarac:
* Upgrades `@lit-labs/react`
* Simplifies logic of react exports
* Simplifies React export types
* Includes JSDoc to React definition files

### DIFF
--- a/scripts/generate-react-exports.js
+++ b/scripts/generate-react-exports.js
@@ -1,15 +1,15 @@
-import fs from 'fs-extra';
-import { format } from 'prettier';
-import { pascalCase } from 'pascal-case';
-import path from 'path';
-import { fileURLToPath } from 'url';
+import fs from "fs-extra";
+import { pascalCase } from "pascal-case";
+import path from "path";
+import { format } from "prettier";
+import { fileURLToPath } from "url";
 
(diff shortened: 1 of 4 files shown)


scripts/generate-react-exports.js
```diff
@@ -73,33 +73,29 @@ for (const module of customElementsModules) {
...
-  const componentLazy = `React.lazy<ReactWebComponent${componentType}>(${componentPromise})`;
-  const componentExport = `export const ${componentName} = ${componentLazy};`;
-
-  baklavaReactFileParts.push(componentExport);
+  const source = `
+  ${jsDoc}
+  export const ${componentName}: import("@lit-labs/react").ReactWebComponent<${componentType}> = (
+    () => createComponent<${componentType}>({
+      react: React,
+      displayName: ${componentName},
```
muratcorlu, line 91: I guess this should be a string?

AykutSarac: Yep, it's a string.

muratcorlu: I mean, it should be `displayName: "${componentName}",`, right?

AykutSarac: Yeah correct, missed it.

PR #710, opened 2023-09-20
feat(input): add showPicker to input

AykutSarac:


src/components/input/bl-input.test.ts
```diff
@@ -44,6 +45,21 @@ describe("bl-input", () => {
...
+  it('should call showPicker if "showPicker" is in HTMLInputElement.prototype', async () => {
+    const el = await fixture<BlInput>(html`<bl-input type="input"></bl-input>`);
+    const spy = stub(el.validationTarget, "showPicker");
+
+    el.showPicker();
+
+    el.requestUpdate();
+    await el.updateComplete;
+
+    expect(HTMLInputElement.prototype).to.have.ownProperty("showPicker");
```
muratcorlu, line 57: Do we need to check this? It's related to the browser support and we have nothing to do if this test fails.

muratcorlu (approved):

PR #713, opened 2023-09-26
feat(input): display type icons

AykutSarac:
I decided that'd be a good idea to display icons depending on the type of the input.

> `icon` attribute overrides the default icon.

image

```html
<bl-input type="search" />
<bl-input type="date" />
<bl-input type="date" icon="academy" />
<bl-input type="time" />
```

Additionally, this PR adds input type of `search`, it could be good idea to review #712 first.


muratcorlu:
I think it will be nice to merge #712 first, then add default icons with this PR

AykutSarac:
> I think it will be nice to merge #712 first, then add default icons with this PR

Actually #712 has no crucial relevance with this PR so this could be reviewed separately.

ogunb:
Should we also update the clear button for type search with this PR? Or should there be another issue for it?

muratcorlu:
> Should we also update the clear button for type search with this PR? Or should there be another issue for it?

I think "clear button for input" is another story. Since some inputs (like search) already puts a native clear button in some browsers, but some others don't. Probably we'll need to make a generic decision that covers all types of inputs. It would be better to consider it separately.

muratcorlu (approved):
It looks good to me.

src/components/input/bl-input.ts
```diff
@@ -284,6 +294,7 @@ export default class BlInput extends FormControlMixin(LitElement) {
 
   firstUpdated() {
     this.setValue(this.value);
+    if (!this.icon) this.icon = inputTypeIcons[this.type];
```
muratcorlu, line 297: I think it's better to have a test for this.

Also some examples in storybook would be helpful to check them visually.

muratcorlu (changes requested):
Unit and UI Test would be great.

buseselvi (approved):

muratcorlu (approved):
Good job! 🚀

fatihhayri (approved):

PR #714, opened 2023-09-28
docs(readme): update cdn example

AykutSarac:


muratcorlu (approved):
We should update this just before making a stable release. Let's be sure to not forget when we release 2.3

PR #716, opened 2023-09-28
fix(react): wrap with React.lazy

AykutSarac:
This is a hotfix for recent generate react scripts change, adds back the forgotten React.lazy. 
Before this change Baklava doesn't work with React 16.

### DIFF
--- a/scripts/generate-react-exports.js
+++ b/scripts/generate-react-exports.js
@@ -28,18 +28,18 @@ for (const module of customElementsModules) {
   const { events, name: componentName, tagName: fileName, jsDoc } = componentDeclaration;
 
   const eventNames = events?.reduce((acc, curr) => {
-    acc[getReactEventName(curr.name)] = curr.name;
-    return acc;
-  }, {}) || {};
+      acc[getReactEventName(curr.name)] = curr.name;
+      return acc;
+    }, {}) || {};
 
   const eventTypes = events?.map(event => {
-    const eventName = getReactEventName(event.name);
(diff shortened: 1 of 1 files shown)


muratcorlu (approved):

muratcorlu (approved):

scripts/generate-react-exports.js
```diff
@@ -52,13 +54,17 @@ for (const module of customElementsModules) {
...
   ${jsDoc}
   export const ${componentName}: React.LazyExoticComponent<ReactWebComponent<${componentType}>> =
-  customElements.whenDefined('${fileName}').then(() => createComponent<${componentType}>({
-    react: React,
-    displayName: "${componentName}",
-    tagName: "${fileName}",
-    elementClass: customElements.get("${fileName}"),
-    events: ${JSON.stringify(eventNames)},
-  }));
+  React.lazy(() =>
```
Enes5519, line 57: How about using generics here instead of assigning a type to 'const'?

muratcorlu: How can we export the component in this way? we need to export a component in a synchronized way.

AykutSarac: We define the return type at line 56.

Enes5519:
This worked in our project. Well done!

scripts/generate-react-exports.js
muratcorlu: Oh sorry, I misunderstood the question.

PR #718, opened 2023-09-28
refactor: default iconPath

0o001:
I believe it could be healthier and more readable this way.
What do you think?


muratcorlu (approved):
Definitely looks better! Thank you!

PR #725, opened 2023-10-01
chore(react): refactor generate-react-exports

AykutSarac:
* Simplifies the logic of `generate-react-exports.js` as types are inferred, removing the need of generics
* Simplifies the React export making it more readable
```tsx
/**
 * @tag bl-textarea
 * @summary Baklava Textarea component
 */
export const BlTextarea = React.lazy(() =>
  customElements.whenDefined("bl-textarea").then(() => ({
    default: createComponent({
      react: React,
      displayName: "BlTextarea",
      tagName: "bl-textarea",
      elementClass: customElements.get("bl-textarea") as Constructor<BlTextarea>,
      events: {
        onBlInput: "bl-input" as EventName<BlTextareaInput>,
        onBlChange: "bl-change" as EventName<BlTextareaChange>,
        onBlInvalid: "bl-invalid" as EventName<BlTextareaInvalid>,
      },
    }),
  }))
);
```


AykutSarac:
`@ts-nocheck` may be removed after #727 resolved.

scripts/generate-react-exports.js
```diff
@@ -24,19 +24,18 @@ const baklavaReactFileParts = [];
...
 
-  const eventNames =
-    events?.reduce((acc, curr) => {
-      acc[getReactEventName(curr.name)] = curr.name;
-      return acc;
-    }, {}) || {};
+  const eventNames = (component.events || []).map(event => {
+    const reactEvent = getReactEventName(event.name);
+      const reactEventName = `${componentName}${reactEvent.split("onBl")[1]}`;
+      return `${reactEvent}: '${event.name}' as EventName<${reactEventName}>`;
```
muratcorlu, line 33: We generate a reference like `EventName<BlTextareaInpu>` but we don't import `BlTextareaInput` anywhere. I think that should cause some issues.

AykutSarac: It's the type of the component so it's referring to self, it shouldn't be a problem because we set BlTextareaInputType to BlTextareaInput above.

AykutSarac: I actually found a way to use the types without repeating the code, just pushed the changes.

AykutSarac: I may have misunderstood your question, the event types are being generated since #711.

scripts/generate-react-exports.js
```diff
@@ -24,43 +24,41 @@ const baklavaReactFileParts = [];
...
-      const predefinedEventName = `${componentName}${eventName.split("onBl")[1]}`;
+  component.events?.forEach(event => {
+    const eventName = getReactEventName(event.name);
+    const eventType = cleanGenericTypes(component.typeParameters, event.type.text);
+    const predefinedEventName = `${componentName}${eventName.split("onBl")[1]}`;
 
-      eventStatements.push(`export declare type ${predefinedEventName} = ${eventType};`);
-      return `${eventName}: EventName<${predefinedEventName}>`;
-    }) || [];
+    eventStatements.push(`export declare type ${predefinedEventName} = ${eventType};`);
```
muratcorlu, line 41: So you just rename `CustomEvent<string>` as `BlTextareaInput` here, right?

AykutSarac: The output of this is
```tsx
export declare type BlTextareaInput = CustomEvent<string>;
```

scripts/generate-react-exports.js
```diff
@@ -24,43 +24,41 @@ const baklavaReactFileParts = [];
...
+  }) || [];
```
Enes5519, line 42: `|| []` is not needed

AykutSarac: Fixed it.

Enes5519 (approved):

PR #732, opened 2023-10-09
feat(icon): review icon is added

buseselvi:


src/components/icon/icon-list.ts
```diff
@@ -110,7 +110,7 @@ const icons = [
   "market_order_cancel",
   "market_order_change",
   "market_order_check",
-  "market_processing",
+  "market_order_processing",
```
AykutSarac, line 113: Are we replacing this? This will cause a breaking change.

AykutSarac (approved):

PR #743, opened 2023-11-22
fix(select): select check additional count event loop

AykutSarac:
Fixes select causing the window to be frozen on choosing x amount of options. This is caused by running `_checkAdditionalItemCount` within updated for each update instead of listening for selected options only.

### DIFF
--- a/src/components/select/bl-select.ts
+++ b/src/components/select/bl-select.ts
@@ -488,7 +488,9 @@ export default class BlSelect<ValueType extends FormValue = string> extends Form
       this.value = null;
     }
 
-    this._checkAdditionalItemCount();
+    if (_changedProperties.has("_selectedOptions")) {
+      this._checkAdditionalItemCount();
+    }
   }
 
   /**


erbilnas (approved):

Enes5519 (approved):

leventozen:
After testing it, I've experienced that the `_checkAdditionalItemCount` method is now running only once per selection, am I missing anythng? Additionally, during the initial load it runs 3 times. We might consider optimizing the `registerOption` method and lifecycles to reduce these initial runs in the future. However, I believe this implementation meets our current needs well. LGTM!
ps: Let's not forget to create an issue for potential future enhancements.

leventozen (approved):
After testing it, I've experienced that the _checkAdditionalItemCount method is now running only once per selection, am I missing anythng? Additionally, during the initial load it runs 3 times. We might consider optimizing the registerOption method and lifecycles to reduce these initial runs in the future. However, I believe this implementation meets our current needs well. LGTM!
ps: Let's not forget to create an issue for potential future enhancements.

AykutSarac:
> After testing it, I've experienced that the _checkAdditionalItemCount method is now running only once per selection, am I missing anythng? Additionally, during the initial load it runs 3 times. We might consider optimizing the registerOption method and lifecycles to reduce these initial runs in the future. However, I believe this implementation meets our current needs well. ps: Let's not forget to create an issue for potential future enhancements.

We could accept this as a workaround, but this is still problematic. It should and only run once when an option is selected/removed, unfortunately rerenders are effecting this and causing to run more than necessary.

leventozen:
> > After testing it, I've experienced that the _checkAdditionalItemCount method is now running only once per selection, am I missing anythng? Additionally, during the initial load it runs 3 times. We might consider optimizing the registerOption method and lifecycles to reduce these initial runs in the future. However, I believe this implementation meets our current needs well. ps: Let's not forget to create an issue for potential future enhancements.
> 
> We could accept this as a workaround, but this is still problematic. It should and only run once when an option is selected/removed, unfortunately rerenders are effecting this and causing to run more than necessary.

As I mentioned, when I test this solution, it runs only once when an option is selected/removed.

Issue #747, opened 2023-12-11
[Feature Request]: Add BlSwitch label


AykutSarac:
### Describe the problem or use case

Switch component needs a label similar to checkbox

### Proposed solution

Accept label as children/slot and set aria-label same as the children

```jsx
<BlSwitch>
  Enable dark mode
</BlSwitch>
```

0o001:
I'm sorry, a lot of tags were created due to my little mistake.

ogunb:
Closed by https://github.com/Trendyol/baklava/pull/754

PR #754, opened 2023-12-23
feat(select): add bl-select label

0o001:
https://github.com/Trendyol/baklava/issues/747

### DIFF
--- a/src/components/switch/bl-switch.css
+++ b/src/components/switch/bl-switch.css
@@ -1,6 +1,7 @@
 :host {
   display: inline-block;
   cursor: pointer;
+  vertical-align: middle;
 }
 
 span {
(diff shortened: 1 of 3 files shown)


src/components/switch/bl-switch.css
```diff
@@ -40,6 +41,32 @@ span {
...
 
+label {
+  display: flex;
+  gap: var(--bl-size-2xs);
+  color: var(--bl-color-neutral-darker);
+  font: var(--bl-font-title-3);
+  cursor: pointer;
+  user-select: none;
+  line-height: normal;
+  align-items: var(--bl-radio-align-items, center);
```
ogunb, line 52: We shouldn't use another components var.
Since we are not allowing user to change flex direction, might as well ditch it.

0o001: Yes, you are right, I will pay attention to this in my other contributions as well. Thank you. I changed it as you said.

leventozen (approved):

PR #756, opened 2023-12-23
fix: EventDispatcher allow events without payload

0o001:


leventozen:
Can you update the rest of the component that sends event with empty string with their tests?

PR #758, opened 2023-12-23
fix(select): return value instead of array on single select

0o001:
https://github.com/Trendyol/baklava/issues/675


erbilnas:
Duplicated PR #677

0o001:
> Duplicated PR #677

You are right, I don't understand how I missed it :). So, I am closing this PR.

PR #759, opened 2023-12-25
feat(select): add search feature for select

### LINKED ISSUE #265: Adding search feature to Select Component

erbilnas:
Closes #265

I've pushed the commit unverified as I'm facing operating system-related test errors on my local machine. Despite my efforts, I couldn't resolve the issue. I've opened an issue on [the relevant project](https://github.com/modernweb-dev/web) for further investigation.

https://github.com/modernweb-dev/web/issues/2589


buseselvi:
Great job! @erbilnas  🚀 I have a few notes:
- The search label does not seem centered.
image

- There should be a divider between search and arrow icons.
image

- We should see the search icon when there is no search in search, and we should only see loading during the search process. When the results are listed, it should return to the search icon after loading.
image

- Also, we shouldn't see 3 icons in any case, we can remove the close button when the search icon appears. The close button can be seen again when the search icon disappears (clicking outside). 🙏 
image

erbilnas:
Hey @buseselvi, thanks for your kind feedback!

> We should see the search icon when there is no search in search, and we should only see loading during the search process. When the results are listed, it should return to the search icon after loading.

For your feedback on this comment, the loading state isn't for typing but for monitoring the fetch process. For instance, if developers aim to display options through an API call, we need to provide them with the loading state. They can then modify the loading state as needed. We talked about it earlier.

fatihhayri (approved):
Ok.

buseselvi:
- can we remove the Select All from the No Data Found result
image

- Also, I think while searching Select All works wrong. It should work as select all searched results, not all results that are not currently visible on the screen. @erbilnas

src/components/select/bl-select.ts
```diff
@@ -245,13 +283,23 @@ export default class BlSelect<ValueType extends FormValue = string> extends Form
   validationTarget: HTMLElement;
 
   open() {
+    if (this.searchBar) {
+      setTimeout(() => {
```
leventozen, line 287: We may consider debouncing in here in the future but rn LGTM.

src/components/select/bl-select.ts
```diff
@@ -325,46 +373,109 @@ export default class BlSelect<ValueType extends FormValue = string> extends Form
...
-    return html`<fieldset
+    const isSearchBarVisible = this.searchBar && this.opened;
+    const isMultipleWithSelection = this.multiple && this._selectedOptions.length > 0;
+
+    const isDividerShown = isSearchBarVisible || isMultipleWithSelection;
+
+    const searchMagIcon = html`<bl-icon
+      class="search-mag-icon"
+      name="search"
+      style="color: var(--bl-color-primary);font-size: var(--bl-font-size-s)"
```
leventozen, line 384: For these, we can consider not using the inline styling. Looks not ugly rn but if we use it everywhere, it will become surely fyi

src/components/select/bl-select.ts
```diff
@@ -453,6 +578,44 @@ export default class BlSelect<ValueType extends FormValue = string> extends Form
...
+      const isVisible = option.textContent?.toLowerCase().includes(this._searchText.toLowerCase());
+
+      option.hidden = !isVisible;
+    });
+
+    this._selectedOptions = this.options.filter(option => option.selected);
+
+    this._handleLastVisibleSearchedOption();
+
+    this.requestUpdate();
```
leventozen, line 602: It is kinda scary but necessary. Let's wait for the feedbacks for the next week till the stable.

leventozen (approved):
I've got some comments but we can take care of them before stable. LGTM! 🚀

PR #755, opened 2023-12-23
fix(select): non label-fixed select breaks input styling

0o001:
https://github.com/Trendyol/baklava/issues/752

PR #760, opened 2023-12-26
fix(select): Using label and placeholder on non label-fixed select br…

fatihhayri:
…eaks input styling

### DIFF
--- a/src/components/select/bl-select.css
+++ b/src/components/select/bl-select.css
@@ -263,8 +263,8 @@ legend span {
   visibility: visible;
 }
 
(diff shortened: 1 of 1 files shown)


erbilnas:
It looks like the duplicated version of #755. Do we need this? We should go through with one MR imho.

erbilnas:
It's looks like the duplicated version of #755. Do we need this or maybe we can go through with

ogunb (approved):

AykutSarac (approved):

PR #761, opened 2023-12-27
feat: add info color and implement it on alert and notification

### LINKED ISSUE #685: Changing the Color of the Informational Component Variants
Informational component variants such as Informational Alerts and Informational Notifications shouldn't be the primary color. Because the primary color is usually the brand color. And brand colors can also be colors such as red, yellow, green. This may cause conflict between the informational and danger/warning/success variants. 

For our default theme, the primary color is the orange color. Orange does not have an informational meaning, and the color we use in the informational alert is very close to the warning variant colors. We did UX research on this topic and users were unable to distinguish between informational and warning variants.

image

Users also match the informational variant with the blue color that we don't have. When we looked for the best practice for our case we saw blue is also known as informational color. 

So, as the UX team, we are suggesting adding **Info Color** to our color palette for informational component variants. And for our default theme, this color should be blue due to research findings. 

image
image

ogunb:
Implement new color: Info.

Uptadate alert and notification components info variant color with the new color and update radius to match design document. 

Also update tooltip component default background color to info. This change can be overridden by `--bl-popover-background-color` variable.

Closes #685

### DIFF
--- a/docs/design-system/colors.stories.mdx
+++ b/docs/design-system/colors.stories.mdx
@@ -29,6 +29,11 @@ Baklava uses a list of defined color with some default values.
     colors={{ '': 'var(--bl-color-warning)', '--highlight': 'var(--bl-color-warning-highlight)', '--contrast': 'var(--bl-color-warning-contrast)' }}
   />
   <ColorItem
+    title="--bl-color-info"
+    subtitle="Info Color"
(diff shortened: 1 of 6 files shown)


fatihhayri (approved):

buseselvi:
Great! The Alert and Notification info variant looks perfect. Can we also make the Tooltip background color `--bl-color-info` 🙏🏻 Because tooltips are also informative components. @ogunb 

image

buseselvi (approved):

AykutSarac (approved):

PR #787, opened 2024-01-29
feat: add localization

### LINKED ISSUE #398: Localization support
For some components, we'll need to provide a solution to set some labels in multi-language. Some examples:

* Month/day names in a calendar or date-picker component.
* Labels of some internal fields. For example [reveal button of password input](#381) or "Go" button in pagination (currently we use attributes for that but it's not practical since you need to set them every time.
* Error messages for inputs. Currently we use native browser error messages and there is no easy way to set them globally

For that kind of very common use cases it's unavoidable to provide a localization solution.

In my opinion requirements are like:

* We should respect HTML's "lang" attribute to set a language.
* It should be easy to set translations globally
* Translations should accept variables or maybe functions
* There should be a solution for pluralization as well

Maybe we can consider using a small library like [shoelace's localize](https://github.com/shoelace-style/localize) that is proven works smoothly and is lightweight enough.

Any opinions are very welcome.

AykutSarac:
* Uses `@lit/localize-tools` and `@lit/localize`
* This adds localization support for Lit components.
* Added storybook document.
* Listens for `html` element's `lang` attribute with mutation observer.
* Not initialized by default, instructions added to storybook.
* Output mode is set to **runtime**, meaning that it has to be capable of changing runtime by rerendering components.
* Added localization support for:
  * Select
  * Pagination

Closes #398


playground/template.html
```diff
@@ -1,37 +1,59 @@
...
+    <bl-select-option value="nl">The Netherlands</bl-select-option>
+    <bl-select-option value="de">Germany</bl-select-option>
+    <bl-select-option value="uk">United Kingdom</bl-select-option>
+  </bl-select>
+
+  <bl-pagination current-page="1" total-items="1500" has-jumper="true">
+  </bl-pagination>
+
+  <bl-pagination current-page="1" total-items="1500" has-select="true">
+  </bl-pagination>
```
erbilnas, line 56: I think we should leave the playground empty

src/generated/locales/tr-TR.ts
```diff
@@ -0,0 +1,19 @@
+
+    // Do not modify this file by hand!
```
erbilnas, line 2: nice warning btw

playground/template.html
erbilnas: > You can try localization within the playground in this branch while the PR is in draft.

I saw this after I commented sorry 😢

src/components/pagination/bl-pagination.test.ts
```diff
@@ -39,8 +39,6 @@ describe("bl-pagination", () => {
     expect(el.itemsPerPage).to.equal(10);
     expect(el.hasJumper).to.equal(false);
     expect(el.hasSelect).to.equal(false);
-    expect(el.jumperLabel).to.equal("Go To");
-    expect(el.selectLabel).to.equal("Show");
```
erbilnas, line 43: Imho instead of removing the tests, we should verify defaults with localized values

src/localization.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const { setLocale } = configureLocalization({
+  sourceLocale,
+  targetLocales,
+  loadLocale: async (locale: string) => localizedTemplates.get(locale) as LocaleModule,
+});
+
+export const init = async () => {
+  const html = document.querySelector("html");
+  const htmlLang = html?.getAttribute("lang") as LangKey | null;
```
erbilnas, line 17: I think we should return void if the lang attribute is not exist

leventozen (approved):

PR #804, opened 2024-02-20
fix(select): do not close the popover if the user presses the space button when searching

### LINKED ISSUE #802: [BUG]: Baklava select input popover close event trigger problem
### Issue description

When typing in the bl select input, pressing the space bar causes the popover to close.

### Media & Screenshots

last

### Baklava Version

_No response_

### Operating system

_No response_

Low (slightly annoying)

erbilnas:
Closes #802


AykutSarac:
What is the root cause of the "Space" causing the popover to be closed here?

erbilnas:
> What is the root cause of the "Space" causing the popover to be closed here?

If you want to leave a blank when searching for something, it triggers keyboard accessibility rules

PR #815, opened 2024-03-25
fix(select): fix select value bug when undefined

### LINKED ISSUE #625: [BUG]: select option empty value
### Issue description

select componentinin altına yazılan option lara eğer value boş girilerse, bu option selected olsa da olmasa da ui a seçili olarak geliyor.

image

### Media & Screenshots

_No response_

### Baklava Version

2.1.0

### Operating system

- OS: [e.g. iOS]:

- Any other details...

High (immediate attention needed)

omerclskn:
This PR fixes the problem of non-valued options appearing selected in select component.

Closes #625

### DIFF
--- a/src/components/select/bl-select.test.ts
+++ b/src/components/select/bl-select.test.ts
@@ -497,6 +497,47 @@ describe("bl-select", () => {
         expect(el.value).to.equal("1");
       });
     });
+    describe("no initial value", () => {
+      it("should not set empty option as selected when multiple and no value", async () => {
+        const el = await fixture<BlSelect>(html`<bl-select multiple name="test">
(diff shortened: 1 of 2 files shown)


AykutSarac (approved):

fatihhayri (approved):

erbilnas:
Sending "empty string" instead of null shoul be better. Otherwise it can break something.

omerclskn:
> Sending "empty string" instead of null shoul be better. Otherwise it can break something.

empty string causes empty values to be set in single select too. my 3rd test in the test file is related to this case. so this bug is also related to single select.

ogunb (approved):

erbilnas (approved):

Issue #816, opened 2024-03-25
[BUG]: <Border of single select clear button is missing>


buseselvi:
### Issue description

I noticed a UI issue in Select Component. In the single selection variant, there should be a border line between the close button and the select arrow icon. It should look like the multi-select variant on the right in the screenshot below.

### Media & Screenshots

<img width="557" alt="image" src="https://github.com/Trendyol/baklava/assets/106681486/e36172e1-5648-493d-8d0c-22c21fe53524">

### Operating system

- OS: [e.g. iOS]:
- Browser [e.g. chrome, safari]:

- Any other details...

### Priority this issue should have

Low (slightly annoying)

### Please review the checkboxes that are applicable.

- [ ] This issue is directly related to React

ogunb:
Fixed by https://github.com/Trendyol/baklava/pull/817

PR #817, opened 2024-03-25
fix(select): fix divider displaying for single variant

ozkersemih:
Bug fix for https://github.com/Trendyol/baklava/issues/816 issue


buseselvi:
The border should not appear when there is no clear button, can you check that? 🙏 
image

erbilnas (approved):

Issue #767, opened 2024-01-13
[Feature Request]: extending drawer width


alpgokcek:
### Describe the problem or use case

When using the Baklava in scenario where i need Drawer, I encountered issues with extending the width of it. I want to make it 80vw but it is set to 424px.

### Proposed solution

We can make it extendible with a variable

leventozen:
what do you think? @buseselvi

buseselvi:
I think we can extend the variants 👌🏻 Maybe we can do it like we did in the Dialog Component, the width of it can depend on the content. Or there are examples of size variants like small, medium, and large for drawers we can do that, too. Let's talk about it in the design meeting. @leventozen @alpgokcek

buseselvi:
We agreed that the Drawer width can be changed by the content, so we can add the free width feature. By default, its width should stay the same (424px). @alpgokcek

PR #820, opened 2024-04-01
feat(drawer): add dynamic width feature to drawer

erdemgonul:
This PR adds setting width feature into drawer.

Closes https://github.com/Trendyol/baklava/issues/767

This had to be written with JavaScript because it was previously written with **@ media**, and we couldn't use CSS variables inside **@ media**. Reference source: https://stackoverflow.com/a/40723269

Also added style-to-px-converter to give width properties like '%', 'vh'

### DIFF
--- a/src/components/drawer/bl-drawer.ts
+++ b/src/components/drawer/bl-drawer.ts
@@ -45,6 +45,12 @@ export default class BlDrawer extends LitElement {
   externalLink?: string;
 
   /**
+   *  Sets the drawer width
+   */
+  @property({ type: String, attribute: "width" })
+  width: string = "424px";
+
+  /**
    * Fires when the drawer is opened
    */
   @event("bl-drawer-open") private onOpen: EventDispatcher<string>;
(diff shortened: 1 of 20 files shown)


buseselvi:
Can we add a drawer example to the storybook with the custom width feature? 🙏🏻

src/utilities/style-to-px.converter.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  let styleInPixel: number | null;
+
+  switch (unit) {
+    case "px":
+      styleInPixel = value;
+      break;
+    case "vh":
+      styleInPixel = (value * window.innerHeight) / 100;
+      break;
```
erbilnas, line 17: Instead of using "vh", I suggest supporting "vw" because it represents a width attribute

src/components/drawer/bl-drawer.ts
```diff
@@ -57,16 +65,42 @@ export default class BlDrawer extends LitElement {
...
     }
+
+    if (changedProperties.has("width")) {
+      this.resizeDrawerWidth();
+    }
   }
 
   private domExistenceSchedule: number;
 
+  private resizeDrawerWidth() {
```
erbilnas, line 92: I don't think we should allow values below a certain threshold

Screenshot 2024-04-02 at 13 44 45

erbilnas (changes requested):

src/components/drawer/bl-drawer.ts
erdemgonul: i added minimum threshold as 100px

src/utilities/style-to-px.converter.ts
erdemgonul: i fixed it, thanks

PR #822, opened 2024-04-03
Calendar component - Calculate and rendering calendar date elements 

dilandoogan:
It calculates days in the selected year and month and shows in calendar


src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,133 @@
...
+  @state()
+  private month: TMonth = MONTHS[new Date().getMonth()];
+
+  @state()
+  private year: number = new Date().getFullYear();
+
+  static get styles(): CSSResultGroup {
+    return [style];
+  }
+  render() {
```
erbilnas, line 44:
```suggestion
  render(): TemplateResult {
```

erbilnas: Additionally, moving the render method to the bottom of the code would increase readability

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,133 @@
...
+    this.month = MONTHS[monthIndex];
+  }
+
+  setYear(year: number) {
+    this.year = year;
+  }
+
+  createCalendarDays() {
+    const monthlyCalendar: Map<number, (number | string)[]> = new Map();
+    let i = 0;
```
erbilnas, line 93: Using meaningful variable names is preferable to using "i"

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,133 @@
...
+    let dayOfTheWeek = 0;
+    let iteratedDay = 1;
+
+    for (; i < this.getDayNumInAMonth() + this.getWeekDayOfDate(); i += 1) {
```
erbilnas, line 97: I believe using the map or forEach functions instead of a basic for loop would enhance readability and performance.

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,133 @@
...
+          })}
+        </div>
+      </div>
+    </div> `;
+  }
+  setMonth(monthIndex: number) {
+    this.month = MONTHS[monthIndex];
+  }
+
+  setYear(year: number) {
```
erbilnas, line 87: We don't need this setter function imho

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,133 @@
...
+  @property({ type: Number, attribute: "start-of-week", reflect: true })
+  startOfWeek?: number = 1;
+
+  @property()
+  locale?: string = document.documentElement.lang;
+
+  @property({ type: Array })
+  disabledDates?: Date | Date[];
+
+  @state()
```
erbilnas, line 32: Defining states with "_" like "_day" is the most common practice in Lit

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,133 @@
...
+  getDayNumInAMonth() {
+    return new Date(this.month.value + 1, this.year, 0).getDate();
+  }
+
+  getWeekDayOfDate() {
+    return new Date(this.year, this.month.value, 1).getDay();
+  }
+
+  setPreviousMonth() {
+    if (this.month.value === 0) {
```
erbilnas, line 122: We can use our MONTHS constant instead of numbers. It would be more readable

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,133 @@
...
+      this.setMonth(MONTHS[11].value);
+      this.setYear(this.year - 1);
+    } else this.setMonth(this.month.value - 1);
+  }
+  setNextMonth() {
+    if (this.month.value === 11) {
```
erbilnas, line 128: here too

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,133 @@
...
+          <bl-button
+            class="arrows"
+            icon="arrow_right"
+            variant="tertiary"
+            kind="neutral"
+            @click="${() => this.setNextMonth()}"
+          ></bl-button>
+        </div>
+        <div class="weekdays">
+          ${Array.from(calendarDays).map(([key, value]) => {
```
erbilnas, line 71: Why would we need the Array prototype in this context?

erbilnas (changes requested):

erbilnas:
Also, please be mindful of our commit rules: https://baklava.design/?path=/docs/documentation-contributing-baklava-commit-rules--documentation

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,133 @@
...
+import style from "./bl-calendar.css";
+import { TCalendarTypes, TMonth } from "./bl-calendar.types";
+
+@customElement("bl-calendar")
+export default class BlCalendar extends LitElement {
+  @property()
+  type: TCalendarTypes = "single";
+
+  @property()
+  minDate?: Date | string;
```
ozkersemih, line 15: I think we should add descriptions for properties like below:
```
/**
*Sets the minimum date value for the calendar
*/
@property()
minDate?: Date | string;
```

src/components/calendar/bl-calendar.ts
dilandoogan: why? do you suggest to set like this.year= this.year-1 in where it is used ?

PR #823, opened 2024-04-04
feat: calendar component

dilandoogan:


scripts/generate-react-exports.js
```diff
@@ -25,6 +25,7 @@ const baklavaReactFileParts = [];
 for (const module of customElementsModules) {
   const { declarations, path } = module;
   const component = declarations.find(declaration => declaration.customElement === true);
+  if(!component) continue;
```
ozkersemih, line 28: when this case may be possible?

dilandoogan: build gives error for calendar's types and constants files so we add this with @Enes5519

ozkersemih: Ok, I understand your concern but I have better idea. `custom-elements-manifest/analyzer` package generates `custom-elements.json` file. You got error because there are modules that created in custom-elements.json file about `bl-calendar.constant.ts` and `bl-calendar.types.ts`. 
When we run `build` command (..`npm run analyze && npm run generate-react-exports`..) first it generates custom-elements.json and generate-react-export script uses that file. So we have `custom-elements-manifest.config.mjs` file to exclude unnecessary files like css, types, constants etc.

```
export default {
  globs: ["src/components/**/!(*.test|*.stories).ts"],
  exclude: ["src/**/*.css","src/components/icon/icon-list.ts"], // we can exclude unnecessary files for custom-elements.json file. We already exclude css files for example.
....
```
We should add pattern for types and constants files. So, final version should be like below:
```
export default {
  globs: ["src/components/**/!(*.test|*.stories).ts"],
  exclude: ["src/**/*.css", "src/**/*.constant.ts","src/**/*/*.types.ts","src/components/icon/icon-list.ts"],
....
```

buseselvi:
Can you add the calendar component stories to the Storybook? UI tests don't work without stories 🙏 @dilandoogan

src/components/calendar/bl-calendar.ts
```diff
@@ -29,14 +30,14 @@ export default class BlCalendar extends LitElement {
...
+  @property({ attribute: false, reflect: true })
+  minDate: Date = new Date(2024, 2, 3);
 
   /**
    * Defines the maximum date value for the calendar
    */
-  @property()
-  maxDate: Date;
+  @property({ attribute: false, reflect: true })
+  maxDate: Date = new Date(2024, 5, 3);
```
ozkersemih, line 40: just to remind, we can give current date as default

AykutSarac:
Good job! Only one thing I noticed is days doesn't have a hover style like in the [design document](https://www.figma.com/design/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=1412-8992&t=8EQ4k8cYAw3jRM5r-4).

buseselvi:
Great job 🚀 I just noticed one thing, can we remove the disabled dates hover? 🙏

src/components/calendar/bl-calendar.css
```diff
@@ -0,0 +1,139 @@
...
+}
+
+.grid-content {
+  display: grid;
+  grid-template-columns: auto auto auto;
+  text-align: center;
+}
+
+.grid-item {
+  width: 93.33px;
```
ozkersemih, line 124: are we really sure to use specific width?

ozkersemih (approved):

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,545 @@
...
+  locale: string = document.documentElement.lang;
+
+  @state()
+  private _selectedDates: CalendarDate[] = [];
+
+  @state()
+  private _selectedRangeDates: RangePickerDates = { startDate: undefined, endDate: undefined };
+
+  @state()
+  private _calendarMonth: number = new Date().getMonth();
```
AykutSarac, line 74: Instead of defining `new Date()` for `_calendarMonth` and `_calendarYear` we can define it once and use.

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,545 @@
...
+      this.setHoverClass();
+    }
+  }
+
+  createCalendarYears() {
+    if (this._calendarYears.length === 0) {
+      for (let i = 4; i >= 0; i--) this._calendarYears.push(this._calendarYear - i);
+      for (let i = 1; i <= 7; i++) this._calendarYears.push(this._calendarYear + i);
+    }
+  }
```
AykutSarac, line 207: That's how I'd refactor it, I recommend finding a better name that explains the function something like `generateSurroundingYears`. 

```suggestion
  createCalendarYears() {
    if (this._calendarYears.length === 0) {
      this._calendarYears = Array.from({ length: 12 }, (_, index) => this._calendarYear - 4 + index);
    }
  }
```

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,545 @@
...
+          this._selectedRangeDates.endDate
+        );
+      }
+    } else if (this._selectedRangeDates.startDate && this._selectedRangeDates.endDate) {
+      this._selectedRangeDates.startDate = calendarDate;
+      this._selectedRangeDates.endDate = undefined;
+      this._selectedDates.splice(0, this._selectedDates.length, this._selectedRangeDates.startDate);
+    }
+    this.setHoverClass();
+  }
```
AykutSarac, line 277: This has excessive cognitive complexity, maybe this could be simpler?

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,545 @@
...
+
+        const endDateParentElement = this.shadowRoot?.getElementById(
+          `${this._selectedRangeDates.endDate?.getTime()}`
+        )?.parentElement;
+
+        endDateParentElement?.classList.add("range-end-day");
+        const rangeDays = [...this.createCalendarDays().values()].flat().filter(
+          date =>
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-expect-error
```
AykutSarac, line 332: Can we remove them and find another way to solve it? Let me know if you need assistance.

AykutSarac (changes requested):

src/components/calendar/bl-calendar.ts
dilandoogan: i dont believe there is , there are four case in there.First if there is not any startDay or endDay,it sets given date as start date.if there is any start date,then there are 2 options.Given date can be bigger than start date,then given date will be end date,if given date smaller than the star date,it should be start date and temp start date should be end date(and start and end date pushed selected dates array).If there are both start date and end date,then they should cleared and start date should be given date

src/components/calendar/bl-calendar.ts
```diff
@@ -0,0 +1,545 @@
...
+  /**
+   *Defines the minimum date value for the calendar
+   */
+  @property({ type: Date, attribute: "min-date", reflect: true })
+  minDate: Date;
+
+  /**
+   * Defines the maximum date value for the calendar
+   */
+  @property({ type: Date, attribute: "max-date", reflect: true })
```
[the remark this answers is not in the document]
dilandoogan: as i tried it doesnt work

src/components/calendar/bl-calendar.ts
dilandoogan: can you check if the solution is good ?

src/components/calendar/bl-calendar.constant.ts
```diff
@@ -1,14 +1,14 @@
...
-export const CALENDAR_VIEWS = {
-  days: "days",
-  months: "months",
-  years: "years",
-} as const;
+export enum CALENDAR_VIEWS {
+  days = "days",
+  months = "months",
+  years = "years",
+}
```
AykutSarac, line 8: Can we use CONSTANT_CASE for enums?
```suggestion
export enum CALENDAR_VIEWS {
  DAYS = "days",
  MONTHS = "months",
  YEARS = "years",
}
```

src/components/calendar/bl-calendar.constant.ts
```diff
@@ -1,14 +1,14 @@
...
 
-export const CALENDAR_TYPES = {
-  single: "single",
-  multiple: "multiple",
-  range: "range",
-} as const;
+export enum CALENDAR_TYPES {
```
AykutSarac, line 10: here too

src/components/calendar/bl-calendar.ts
```diff
@@ -70,11 +69,13 @@ export default class BlCalendar extends LitElement {
   @state()
   private _selectedRangeDates: RangePickerDates = { startDate: undefined, endDate: undefined };
 
+  today = new Date();
```
AykutSarac, line 72: Should we mark this as private?

AykutSarac (changes requested):

AykutSarac (approved):

AykutSarac (dismissed):

ozkersemih (approved):

ozkersemih (approved):

AykutSarac (approved):

Issue #830, opened 2024-04-17
[Feature Request]: Select search should support international character set


ogunb:
### Describe the problem or use case

Currenly user cannot search with Turkish letters:

<img width="489" alt="Screenshot 2024-04-17 at 15 12 51" src="https://github.com/Trendyol/baklava/assets/24693913/7ae46421-4ce3-49c4-b9e1-92d710a39aba">
<img width="494" alt="Screenshot 2024-04-17 at 15 13 02" src="https://github.com/Trendyol/baklava/assets/24693913/832342ee-7419-4679-8ee0-1467bc623a12">

### Proposed solution

We could use `localeCompare` or `toLocaleLowerCase` with html lang tag.

### Additional technical context

I think we should also support not using those letters. Ex: İçecekler should be filtered with both `İç` and `ıc`. This would require more work than simply using locale functions though. Maybe we can look into some lightweight search libraries?

PR #843, opened 2024-04-24
feat(select): search locale support for  #830

cenarturkmen:
I added toLocaleLowerCase method for the search function of select. 
Also created a util method for getting locale. I'm not sure why we are not exporting getLocale from "configureLocalization"


cenarturkmen:
I couldn't link but issue: [830](https://github.com/Trendyol/baklava/issues/830)

src/localization.ts
```diff
@@ -14,11 +15,9 @@ const { setLocale } = configureLocalization({
 
 export const init = async () => {
   const html = document.querySelector("html");
```
erbilnas, line 17: I believe we can do away with this line now

src/localization.ts
```diff
@@ -1,6 +1,7 @@
 import { LocaleModule, configureLocalization } from "@lit/localize";
 import { sourceLocale, targetLocales, allLocales } from "./generated/locale-codes";
 import * as templatesTR from "./generated/locales/tr";
+import { getLocale } from "./utilities/getLocale";
 
 type LangKey = "tr" | "en";
```
erbilnas, line 6: this one too

erbilnas (changes requested):

src/utilities/getLocale.ts
```diff
@@ -0,0 +1,14 @@
+import { allLocales } from "../generated/locale-codes";
+
+type LangKey = "tr" | "en";
+
+export const getLocale = (): LangKey => {
```
ogunb, line 5: `configureLocalization` exposes a getLocale method. We can export that from localization.ts and use that instead. Issue detail might have been misleading sorry about that.

ogunb (changes requested):

ogunb:
Also, would you be able to document this behavior in Storybook?

Thanks for your contribution btw

ozkersemih (approved):

buseselvi:
Closed this as completed in [#857](https://github.com/Trendyol/baklava/pull/857). If there is something missing, we can open an issue separately. Thank you 🙏 @cenarturkmen

PR #844, opened 2024-04-25
fix(select): fix recurring space closing select box issue

### LINKED ISSUE #802: [BUG]: Baklava select input popover close event trigger problem

ogunb:
Keydown is triggered before input,
causing the select keydown handler to run first.

Fixes #802

### DIFF
--- a/src/components/select/bl-select.ts
+++ b/src/components/select/bl-select.ts
@@ -432,6 +432,11 @@ export default class BlSelect<ValueType extends FormValue = string> extends Form
                 ? searchbarPlaceholderText
                 : this.label || searchbarPlaceholderText}
               @input=${this._handleSearchOptions}
+              @keydown=${(e: KeyboardEvent) => {
+                if (e.code === "Space") {
+                  e.stopPropagation();
+                }
+              }}
               .value=${this._searchText}
             />
           `}
--- a/src/components/select/bl-select.test.ts
+++ b/src/components/select/bl-select.test.ts
@@ -423,6 +423,26 @@ describe("bl-select", () => {
     await(()=> expect(document.activeElement).to.equal(el));
   });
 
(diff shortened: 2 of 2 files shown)


buseselvi (approved):

src/components/select/bl-select.ts
```diff
@@ -432,6 +432,11 @@ export default class BlSelect<ValueType extends FormValue = string> extends Form
                 ? searchbarPlaceholderText
                 : this.label || searchbarPlaceholderText}
               @input=${this._handleSearchOptions}
+              @keydown=${(e: KeyboardEvent) => {
+                if (e.code === "Space") {
+                  e.stopPropagation();
+                }
+              }}
```
ozkersemih, line 439: on line 576, there is a code block below. We can achieve same behaviour by updating that part of bl-select component.
```typescript
  private handleKeydown(event: KeyboardEvent) {
    if (this.focusedOptionIndex === -1 && ["Enter", "Space"].includes(event.code)) {
      this._togglePopover();
      event.preventDefault();
    }
...
```


if we don't to close popover when user press `Space`, we can achieve this behaviour by removing space in that array `(["Enter", "Space"])` in condition. Because probably we wanted to close popover our own, if now we dont want, we should update that condition I think. What do you think?

ogunb: Not closing the popover if nothing is focused makes sense to me too. @buseselvi what do you say?

ozkersemih (approved):

src/components/select/bl-select.ts
ogunb: We'll discuss this further later.

PR #847, opened 2024-04-29
fix(dialog): fix nested dialog propagation problem

### LINKED ISSUE #342: Bubbling and propagation problem in nested dialog
**Describe the bug**
When using nested dialog in the Vue 2 project, we noticed that every dialog closes when we click the close button of the child dialog or press the esc key.

**To Reproduce**
1. Go to https://codesandbox.io/s/vue-2-playground-forked-wrxmzy?file=/src/App.vue
2. Click on open parent dialog button
3. Click on open child dialog button
4. Click on close button
5. You will see the every dialog is closed

**Expected behavior**
When we click the close button of child dialog or esc key we should only see it close.

 - OS: Macos
 - Browser Chrome

**Additional context**
This problem probably occur because of bubbling and propagation. When I add the stop (it is equal event.stopPropagation()) like this:

```html
<bl-dialog
 :open="openChildDialog"
 @bl-dialog-close.stop="openChildDialog = false">
   Child Dialog
</bl-dialog>
```

The `stop` fixes the close button problem but esc problem is going on.

BaharYilmaz:
This PR fixes the propagation issue during close in nested dialogs

Closes #342

### DIFF
--- a/src/components/dialog/bl-dialog.test.ts
+++ b/src/components/dialog/bl-dialog.test.ts
@@ -208,6 +208,26 @@ describe("bl-dialog", () => {
       await resetMouse();
     });
 
+    it("should prevent parent dialog from closing when the child dialog is closed", async () => {
+      const dialog = await fixture<typeOfBlDialog>(html`<bl-dialog id="parent" open>
+        <bl-dialog id="child" open>
+          Child dialog
+        </bl-dialog>
+      </bl-dialog>`);
+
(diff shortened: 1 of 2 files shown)


AykutSarac (approved):

erbilnas (approved):

ogunb (approved):

ozkersemih:
It is merged but i think this approach is not best solution because now i can not listen bl-dialog-close event like:

```
  window.addEventListener('bl-dialog-close',(e) => console.log(e));
```

Issue #391, opened 2023-01-25
Masking Capability for Input Component


muratcorlu:
Leading and trailing texts (like currencies or units) would be possible to put our Input component. Details will come later.

muratcorlu:
Design changes are ready: https://www.figma.com/file/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=4%3A5586&t=dpos9bPGKoCkqo5m-0

doganozturk:
This task had originally title **Leading/Trailing text for Input field**.

We talked about this and decided to implement this feature with usage of input's placeholder functionality and usage of input masking.

doganozturk:
Our latest decision is not to implement input masking as a feature but provide detailed examples with compatible libraries. **Leading/Trailing text for Input field** should be achieved by placeholder usage and input masking integration with 3rd party libraries such as IMask and Maskito.

PR #854, opened 2024-05-03
story(input): add input masking documentation

doganozturk:
Closes [#391](https://github.com/Trendyol/baklava/issues/391)

### DIFF
--- a/src/components/input/bl-input.stories.mdx
+++ b/src/components/input/bl-input.stories.mdx
@@ -361,6 +361,114 @@ When you run this example and submit the form, you'll see key/value pairs of the
 
 <bl-alert icon>If user presses `Enter` key in an input inside a form, this will trigger submit of the form. This behaviour mimics the native input behaviour.</bl-alert>
 
+## Input Masking
+
+Input masking is essential in user interfaces that require formatted entries such as phone numbers and dates, ensuring consistent and structured user input.
+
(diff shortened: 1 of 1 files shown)


src/components/input/bl-input.stories.mdx
```diff
@@ -361,6 +361,114 @@ When you run this example and submit the form, you'll see key/value pairs of the
...
+
+      return {
+        blInputRef,
+      };
+    },
+  });
+</script>
+```
+
+In both examples, the `useIMaskWithObserver` hook is a custom hook that combines IMask with a MutationObserver to ensure that the masking behavior is applied correctly and consistently. Respective example implementations for such an approach can be found in [here (React)](https://github.com/doganozturk/baklava-input-mask-react-examples/blob/f9f2155d0c7be477fa6e87ae03a8c5c9ac1a27df/src/examples/imask/hooks/useIMaskWithObserver.js) and [here (Vue)](https://github.com/doganozturk/baklava-input-mask-vue-examples/blob/9342e453c6d66525904dfa01c68fa09934c066e5/src/examples/imask/hooks/useIMaskWithObserver.js).
```
Enes5519, line 462: It didn't feel right to share personal links. If necessary, we can open a folder called `examples` in the repo.

doganozturk: I though about it, too. Is it possible for us to serve different projects through our public domain and without too much effort? This projects have entirely different setups and that's the point. We should talk about in our weekly 👍

PR #872, opened 2024-05-15
style(dialog): tab-group inside dialog

### LINKED ISSUE #544: Using tab component inside dialog

ozkersemih:
closes #544


AykutSarac (approved):
Good job! I noticed that when the dialog doesn't have a title and is critical there's an unnecessary margin at the top coming from the `section.content` we should investigate on it later.

image

image

AykutSarac (approved):

erbilnas (approved):

Issue #873, opened 2024-05-16
[BUG]: Disabled select option can be selected with enter


gokcecicek:
### Issue description

While navigating the select options with the arrow keys, a disabled option can be selected with enter. If the option is disabled, it should not be selected.

### Media & Screenshots

<img width="242" alt="Screenshot 2024-05-16 at 14 43 13" src="https://github.com/Trendyol/baklava/assets/52400850/01bbb82d-be7e-4072-806a-9a6699197010">

<img width="242" alt="Screenshot 2024-05-16 at 14 43 28" src="https://github.com/Trendyol/baklava/assets/52400850/77baded2-e563-4630-8e47-a0811f1108ca">

### Operating system

- OS: [e.g. iOS]: MacOS
- Browser [e.g. chrome, safari]: Chrome, Firefox, Safari

- Any other details...

### Priority this issue should have

Medium (should be fixed soon)

### Please review the checkboxes that are applicable.

- [ ] This issue is directly related to React

AykutSarac:
Disabled option shouldn't be focused. @ahmetozantekin Would it be possible for you to updating your PR that way?
fyi @Trendyol/baklava

ahmetozantekin:
hi, @AykutSarac it's already like that. In [L. 615](https://github.com/Trendyol/baklava/pull/874/files#diff-84e8f4fecfe0b3502966742219aeb6ad82b25894951e1943f4f39a9f780f765aR615) filtered the non-disabled items. ArrowDown, ArrowUp keys and focus function work with that filtered (only active) items.

https://github.com/Trendyol/baklava/assets/3971877/91c262b4-4f85-4cb0-b21e-16625ef479c7

PR #874, opened 2024-05-16
fix(select): fix disabled select option can be selected with enter

ahmetozantekin:
This pr fixes navigating in the select options with the arrow keys, disabled option can be selected with enter bug.

Closes [873](https://github.com/Trendyol/baklava/issues/873)

### DIFF
--- a/playground/template.html
+++ b/playground/template.html
@@ -8,7 +8,7 @@
     <script type="module" src="./dist/baklava.js"></script>
     <script>
       // Live reload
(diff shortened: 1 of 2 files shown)


AykutSarac (approved):

leventozen (approved):

buseselvi (approved):

buseselvi:
Since there are bl-checkbox components in the multiple select options, it cannot be selected with enter, but let's open this as a separate issue.

PR #875, opened 2024-05-17
fix(drawer): fix drawer delay bug for removing from dom

### LINKED ISSUE #536: Drawer has delay for removing from dom
**Description**
When user closes drawer, there is 1000ms delay for animation. This also causes delay for removing operation from dom. We can solve it with check animation ended or not with more suitable method.

**Expected behavior**
We should not use timeout when we remove component from dom. We should check with function.
**Screenshots**
image

fatihkucuktry:
Closes #536


doganozturk (approved):
I'm not entirely sure about exposing ﻿`--bl-drawer-animation-duration`, but this change resolves the issue. We might need to get another opinion on this.

ozkersemih (approved):

doganozturk (approved):

leventozen (approved):

PR #877, opened 2024-05-21
feat: spinner component

### LINKED ISSUE #866: [Component]: <Spinner Component>
## Figma Design Document

[Spinner Design Document](https://www.figma.com/design/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=22561%3A4806&t=fD181faLQJunYb5c-1)

## Implementation

General usage example:

```html
<bl-spinner></bl-spinner>
```

### Rules

- Spinner can be used to indicate loading states.
- Custom sizes and colors can be applied using CSS variables or inline styles.
- The spinner can be displayed in overlay mode to cover its parent element.

### Usage Examples

Default size is `var(--bl-font-size-m)` and default color is `var(--bl-color-primary)`. But any size and color can be set like this:

```html
<bl-spinner size="32px" color="#FFF"></bl-spinner>
```

The color and size can be set like this:

```html
<bl-spinner size="3rem" color="var(--bl-color-success)"></bl-spinner>
```

The disabled can be set like this:

```html
<bl-spinner disabled></bl-spinner>
```

The overlay can be set like this:

```html
<bl-spinner overlay size="var(--bl-font-size-l)" color="var(--bl-color-secondary)"></bl-spinner>
```

### Attributes

| Attribute | Description                                 | Default Value           |
|-----------|---------------------------------------------|-------------------------|
| `size`    | Size of the spinner (e.g., `3rem`, `32px`, `var(--bl-font-size-m)`) | `var(--bl-font-size-m)` |
| `color`   | Color of the spinner                        | `var(--bl-color-primary)`|
| `disabled`| Whether the spinner is disabled             | `false`                 |
| `overlay` | Whether the spinner is in overlay mode      | `false`                 |

MertOzbudak:
closes #866

### DIFF
--- a/src/baklava.ts
+++ b/src/baklava.ts
@@ -25,6 +25,7 @@ export { default as BlDropdown } from "./components/dropdown/bl-dropdown";
 export { default as BlDropdownItem } from "./components/dropdown/item/bl-dropdown-item";
 export { default as BlDropdownGroup } from "./components/dropdown/group/bl-dropdown-group";
 export { default as BlSwitch } from "./components/switch/bl-switch";
+export { default as BlSpinner } from "./components/spinner/bl-spinner";
 export { default as BlNotification } from "./components/notification/bl-notification";
 export { default as BlNotificationCard } from "./components/notification/card/bl-notification-card";
 export { default as BlTable } from "./components/table/bl-table";
--- a/src/components/button/bl-button.css
+++ b/src/components/button/bl-button.css
@@ -1,13 +1,3 @@
-@keyframes spin {
-  from {
-    transform: rotate(0deg);
-  }
-
-  to {
-    transform: rotate(359deg);
-  }
-}
-
 :host {
   display: var(--bl-button-display, inline-block);
   max-width: 100%;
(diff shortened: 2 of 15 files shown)


ozkersemih (dismissed):

doganozturk (dismissed):

AykutSarac (commented):
I'm not sure if we should be showing loading cursor on hover, wouldn't it be better to display the default instead?
@buseselvi what do you think?

Also a small thing but can we increase the spinner size for the *Relative Overlay* story?

MertOzbudak:
> I'm not sure if we should be showing loading cursor on hover, wouldn't it be better to display the default instead? @buseselvi what do you think?
> 
> Also a small thing but can we increase the spinner size for the _Relative Overlay_ story?


We discussed this in the meeting as well. The size is adjustable via a prop. Is there another concern here? @AykutSarac

AykutSarac (approved):

doganozturk (approved):

ozkersemih (approved):

PR #902, opened 2024-08-05
feat(dialog): add width and caption line clamp css property

### LINKED ISSUE #737: [Feature Request]: Dialog title is too long

The Dialog caption is currently **nowrap**, but this can be ugly in some cases and it may be necessary to increase it to 2 lines.

**Actual**
image

**Expected**
image

### Proposed solution

We can add a line clamp and keep its value in a css variable.

https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp

### Alternatives considered

We can add the ability to adjust the dialog width. We can also remove Nowrap from the caption.

_No response_

Enes5519:
Added the --bl-dialog-width and --bl-dialog-caption-line-clamp CSS properties.

Closes #737

### DIFF
--- a/src/components/dialog/bl-dialog.css
+++ b/src/components/dialog/bl-dialog.css
@@ -8,6 +8,7 @@
   display: flex;
   flex-direction: column;
   background: var(--background-color);
+  width: var(--bl-dialog-width, auto);
   max-width: calc(100vw - var(--bl-size-4xl));
   max-height: calc(100vh - var(--bl-size-4xl));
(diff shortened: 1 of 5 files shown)


buseselvi:
Hi, I just noticed that the top padding of the close button seems wider than it should be, can you check? 🙏 @Enes5519 

image

erbilnas (approved):

erbilnas (approved):

AykutSarac (approved):

PR #909, opened 2024-09-05
fix(tooltip): use mouseenter as event listener

### LINKED ISSUE #908: [BUG]: bl-tooltip flickering
### Issue description

1- if bl-tooltip has more than once child
2- behaviour of tooltip starts to become weird
3- sometimes it is displaying, sometimes not

This issue happens when we use React but we are not sure it is directly related to React or not.

### Media & Screenshots

https://github.com/user-attachments/assets/14a4ff27-6f0f-4aaf-9ea6-46f826f24dc7

### Baklava Version

2.3.0-beta.7

### Operating system

- OS: [e.g. iOS]:

- Any other details...

Low (slightly annoying)

ozkersemih:
closes #908 

I noticed there is a problem when we use `mouseover` as event listener in bl-tooltip component. mouseever triggered for every children inside tooltip trigger slot. It should be triggered only once when mouse entered the tooltip trigger slot element instead of its children. If we use `mouseenter` as event listener it seems problem solved.

I put the video about this case also:

https://github.com/user-attachments/assets/a7b4ea3b-8c5a-47e9-9dbc-69a62da5d7d1

### DIFF
--- a/src/components/tooltip/bl-tooltip.ts
+++ b/src/components/tooltip/bl-tooltip.ts
@@ -61,7 +61,7 @@ export default class BlTooltip extends LitElement {
 
     if (target) {
       target.addEventListener("focus", this.show, { capture: true });
-      target.addEventListener("mouseover", this.show);
+      target.addEventListener("mouseenter", this.show);
       target.addEventListener("blur", this.hide, { capture: true });
       target.addEventListener("mouseleave", this.hide);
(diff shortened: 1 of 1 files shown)


alibalbars:
i also had this problem. nice catch!

doganozturk (approved):

PR #940, opened 2024-10-15
feat(calendar): Enhance accessibility by adding visually hidden text …

### LINKED ISSUE #939: [Feature Request]: <Enhance Calendar Accessibility>

issues with the calendar navigation buttons, where screen readers do not properly recognize them. The aim of this issue is to address this concern by implementing a change.

### Proposed solution

Add Visually Hidden Text to Calendar Navigation Buttons

### Alternatives considered

_No response_

The visually hidden text has been implemented using a span element with a class that applies CSS rules to hide it from sighted users while keeping it accessible to screen readers.

mryunt02:
### Changes
- Added visually hidden text to the navigation buttons for improved accessibility:
  - `<span class="visually-hidden">Previous Calendar View</span>` was added to the previous button.
  - `<span class="visually-hidden">Next Calendar View</span>` was added to the next button.

### CSS
- Added the following CSS class to hide the text visually while keeping it accessible to screen readers:
  ```css
  .visually-hidden {
    position: absolute;
    width: 1px;
    height: 1px;
    margin: -1px;
    padding: 0;
    overflow: hidden;
    clip: rect(0, 0, 0, 0);
    border: 0;
  }
### Related Issue
Fixes #939


erbilnas (changes requested):
Thank you so much for this finding about accessibility 🙏 Setting the `label` attribute for bl-button works simply fine—no need to add an extra class. Example: [Icon-Only Buttons](https://baklava.design/?path=/docs/components-button--documentation#icon-only-buttons)

src/components/calendar/bl-calendar.ts
```diff
@@ -521,7 +521,9 @@ export default class BlCalendar extends LitElement {
             variant="tertiary"
```
erbilnas, line 521: label="Previous Calendar View" for example

erbilnas:
Also, if you could apply the same for other buttons, we’d really appreciate it! 🧡

src/components/calendar/bl-calendar.ts
mryunt02: Oh! I missed the label property of button elements, Thank you. I'm on it

erbilnas (approved):

src/components/calendar/bl-calendar.ts
AykutSarac: I’d suggest updating it to “next/previous year” if possible, as the buttons specifically navigate between **years**. The current version works as well, though.

mryunt02: I thought about changing the button label to next/previous year, but can also switch between months with this button. So I thought the most logical is the previous/next calendar view.

AykutSarac: I think we can make it dynamic in this case depending on the use and state.

mryunt02: Yes, this make sense. I'm on it.

src/components/calendar/bl-calendar.ts
```diff
@@ -511,13 +511,14 @@ export default class BlCalendar extends LitElement {
     const showMonthSelected =
       this._calendarView === CALENDAR_VIEWS.MONTHS ? "header-text-hover" : "";
     const showYearSelected = this._calendarView === CALENDAR_VIEWS.YEARS ? "header-text-hover" : "";
+    const ButtonLabel = this._calendarView === CALENDAR_VIEWS.DAYS ? "Month" : "Year";
```
erbilnas, line 514: the variables should be camelCase

erbilnas (approved):

AykutSarac (approved):

PR #946, opened 2024-10-18
feat(checkbox): Add keyboard handling for Enter and Space keys to tog…

mryunt02:
### Summary of Changes
- Added keyboard functionality to the input component to toggle the checked state when the Enter or Space key is pressed.
### Changes Made
1. Implemented the handleKeyDown function:
```
private handleKeyDown(event: KeyboardEvent) {
    if (event.code === "Enter" || event.code === "Space") {
        this.checked = !this.checked;
        this.onChange(this.checked);
        event.preventDefault();
    }
}

```
2. Associated the function with the @keydown event listener:
`<input @keydown=${this.handleKeyDown} />`

### Reason for Changes
This change was made to enhance user interaction with form elements via the keyboard and to improve accessibility.


src/components/checkbox-group/checkbox/bl-checkbox.ts
```diff
@@ -177,6 +177,13 @@ export default class BlCheckbox extends FormControlMixin(LitElement) {
   private handleFieldValueChange = (event: CustomEvent<Array<string>>) => {
     this.checked = event.detail.includes(this.value);
   };
+  private handleKeyDown(event: KeyboardEvent) {
+    if (event.code === "Enter" || event.code === "Space") {
```
erbilnas, line 181: I believe we should avoid using the Enter key for checking

mryunt02: Yes, thank you. Enter key is not supported in checkbox accessibility. I missed that part.

Enes5519 (approved):

Issue #950, opened 2024-10-21
[Feature Request]: <Enhance Radio Group Accessibility>


mryunt02:
### Describe the problem or use case

For best accessibility and user experience, the selected Radio Group should be toggled with the arrow keys.

**Arrow Keys:**
- **Down/Right Arrow:**
  - Moves focus to and checks the next radio button in the group.
  - If focus is on the last radio button, moves focus to the first radio button.
  - The state of the previously checked radio button is changed to unchecked.
- **Up/Left Arrow:**
  - Moves focus to and checks the previous radio button in the group.
  - If focus is on the first radio button, moves focus to and checks the last radio button.
  - The state of the previously checked radio button is changed to unchecked.

Additionally:
- The Radio Button should not have the `aria-readonly` property.
- Each Radio Button should have the `aria-checked` attribute to indicate its checked state.

Reference: [W3C ARIA Radio Group Example](https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/)

PR #951, opened 2024-10-21
feat(radio, radio-group): Enhance accessibility with arrow key navigation and aria attributes

mryunt02:
This PR addresses and resolves issue [#950](https://github.com/Trendyol/baklava/issues/950).

**Changes Made:**
- Implemented arrow key navigation for the Radio Group component.
  - **Down/Right Arrow:**
    - Moves focus to and checks the next radio button in the group.
    - If focus is on the last radio button, moves focus to the first radio button.
    - The state of the previously checked radio button is changed to unchecked.
  - **Up/Left Arrow:**
    - Moves focus to and checks the previous radio button in the group.
    - If focus is on the first radio button, moves focus to and checks the last radio button.
    - The state of the previously checked radio button is changed to unchecked.
- Removed the `aria-readonly` property from the Radio Button component.
- Added the `aria-checked` attribute to the Radio Button component to indicate its checked state.

These changes improve the accessibility and user experience of the Radio Group component in compliance with [W3C ARIA Radio Group Example](https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/).

This PR resolves issue #950.


mryunt02:
this PR fails the test but I think it is necessary, I did not change the test file to avoid breaking it.

src/components/radio-group/radio/bl-radio.test.ts
```diff
@@ -19,7 +19,7 @@ describe("bl-radio", () => {
       `<div
         aria-disabled="false"
         aria-labelledby="label"
-        aria-readonly="false"
```
erbilnas, line 22: I don’t quite understand why you did this

mryunt02: aria-readonly property is not supported in Radio Group according to [W3C ARIA Radio Group Example](https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/)

Enes5519 (changes requested):
It would be better to capitalize the first letter of the labels.

Enes5519 (approved):

erbilnas (approved):

PR #955, opened 2024-11-06
feat(tag-component): tag component

dilandoogan:
## Figma Design Document
https://www.figma.com/design/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=25001-1763&t=xA7DTWjCkA2EqaDc-1 

## Implementation

General usage example:

```html
<bl-tag>In Progress</bl-tag>
```

### Usage Examples
Selectable variant usage:
```html
<bl-tag @bl-tag-click="handleTagClick" selected>Selectable tag</bl-tag>
```
The removable variant can be set like this:

```js

const handleTagClick=(event)=>{
   tags.filter((tag)=>tag.value!==event.value)
}
<bl-tag variant="removable" @bl-tag-click="handleTagClick">Removable tag</bl-tag>
```

The icon can be set like this:

```html
<bl-tag icon="info">Default</bl-tag>
```

The size and disabled attributes can be set like this:

```html
<bl-tag size="large" disabled>In Progress</bl-tag>
```

#### Slots

| Name        | Description     | Default Content |
|-------------|-----------------| --------------- |
| `icon` slot | Icon of the tag | - |

#### Attributes

| Attribute            | Description                                   | Default Value |
|----------------------|-----------------------------------------------|---------------|
| size (`string`)      | Size of tag(`small`,`medium`,`large`)         | medium        |
| icon (`bl-icon`)     | Name of the icon that will be shown in tag    | -             |
| variant (`string`)   | Variants of the tag(`selectable`,`removable`) | selectable    |
| disabled (`boolean`) | Makes tag disabled                            | false         |
| selected (`boolean`) | Makes tag selected                            | false         |
| value (`string`)     | Sets tags value                               | -             |

### Events

| Name           | Description                | Payload                           |
|----------------|----------------------------|-----------------------------------|
| `bl-tag-click` | Fires when tag is clicked  | `{value:string,selected:boolean}` |


Enes5519 (dismissed):

fatihhayri (approved):

erbilnas (changes requested):
In my opinion we should also add a tag-group wrapper to get all selected tags' ids. Also using bl-click instead of bl-tag-click is enough I guess

AykutSarac:
> In my opinion we should also add a tag-group wrapper to get all selected tags' ids. Also using bl-click instead of bl-tag-click is enough I guess

Yes, that's in the roadmap but we decided to develop the tag first and group element later.

src/components/tag/bl-tag.css
```diff
@@ -0,0 +1,99 @@
...
+  padding-block: var(--padding-vertical);
+  padding-inline: var(--padding-horizontal);
+  background-color: var(--bg-color);
+  color: var(--color, white);
+  font: var(--font);
+  font-kerning: none;
+  height: var(--height);
+}
+
+:host([variant="selectable"]) .tag {
```
erbilnas, line 35: Instead of using :host every time, I think we can use the .tag class

src/components/tag/bl-tag.ts
```diff
@@ -0,0 +1,95 @@
...
+  static get styles(): CSSResultGroup {
+    return [style];
+  }
+
+  @query(".remove-button") removeButton!: HTMLButtonElement;
+
+  /**
+   * Sets the tag size
+   */
+  @property({ type: String, reflect: true })
```
erbilnas, line 29: Can we use TagSize and TagVariant as types on decorators instead?

src/components/tag/bl-tag.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  render(): TemplateResult {
+    const icon = this.icon ? html`<bl-icon name=${this.icon}></bl-icon>` : "";
+
+    const removeButton =
+      this.variant === "removeable"
+        ? html`
+            <bl-button
+              icon="close"
+              size=${this.size === "large" ? "medium" : "small"}
```
erbilnas, line 68: I don’t fully understand that, but perhaps we can separate it as a computed property

src/components/tag/bl-tag.ts
AykutSarac: we can't. TagVariant and TagSize are TypeScript types, you should pass JavaScript definitions here.

src/components/tag/bl-tag.ts
```diff
@@ -0,0 +1,95 @@
...
+              class="remove-button"
+              ?disabled=${this.disabled}
+              @bl-click=${this.handleClick}
+            ></bl-button>
+          `
+        : "";
+
+    return html`<div
+      class="tag"
+      @click=${this.variant === "selectable" ? this.handleClick : undefined}
```
erbilnas, line 81: ifDefined will be better I guess

src/components/tag/bl-tag.css
AykutSarac: that's how we used selectors at other elements as well to target e.g. specific variant like "selectable"

src/components/tag/bl-tag.ts
AykutSarac: The 'ifDefined' directive has no effect here. - says TypeScript.
Click for selectable is not working already at the moment so I'll dive deeper here later.

erbilnas (dismissed):

src/components/calendar/bl-calendar.test.ts
```diff
@@ -534,7 +534,7 @@ describe("bl-calendar", () => {
 
     const setNextCalendarViewSpy = sinon.spy(element, "setNextCalendarView");
 
-    element.handleDate(new Date(2024, 2, 15));
+    element.handleDate(new Date(element.today.getFullYear()+1, 2, 15));
```
AykutSarac, line 537: I think we should handle such bug fixes at separate PR, it'll hard to track past changes through commit messages as they'll be merged into one, PR title

dilandoogan: okey i will seperate it to another pr

Enes5519 (approved):

ozkersemih (approved):

buseselvi:
There have been changes in the paddings in the design, can you check them? 🙏

AykutSarac (dismissed):

fatihhayri (dismissed):

ogunb (approved):

PR #962, opened 2024-11-28
refactor: improve form validation

Enes5519:
customInvalidText, forceCustomError and clearCustomError are deprecated

add error attribute

### DIFF
--- a/src/components/input/bl-input.ts
+++ b/src/components/input/bl-input.ts
@@ -335,7 +335,7 @@ export default class BlInput extends FormControlMixin(LitElement) {
       this.requestUpdate();
     }
 
-    if (changedProperties.has("error")) {
+    if (changedProperties.has("error") && this.error && !this.dirty) {
       this.reportValidity();
     }
   }
--- a/src/components/input/bl-input.ts
+++ b/src/components/input/bl-input.ts
@@ -3,6 +3,7 @@ import { customElement, property, query, state } from "lit/decorators.js";
 import { classMap } from "lit/directives/class-map.js";
 import { ifDefined } from "lit/directives/if-defined.js";
 import { live } from "lit/directives/live.js";
+import { localized, msg } from "@lit/localize";
 import { FormControlMixin } from "@open-wc/form-control";
 import { submit } from "@open-wc/form-helpers";
 import "element-internals-polyfill";
@@ -45,6 +46,7 @@ export type InputSize = "small" | "medium" | "large";
  * @cssproperty [--bl-input-padding-end] Sets the padding end
  */
 @customElement("bl-input")
+@localized()
 export default class BlInput extends FormControlMixin(LitElement) {
   static get styles(): CSSResultGroup {
     return [style];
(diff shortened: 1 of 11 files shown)


doganozturk (dismissed):

src/components/input/bl-input.ts
```diff
@@ -184,17 +184,24 @@ export default class BlInput extends FormControlMixin(LitElement) {
...
 
+  /**
+   * @deprecated
+   */
   get customInvalidText(): string {
     return this._customInvalidText;
   }
 
+  @property({ reflect: true, type: String })
+  error: string = "";
```
AykutSarac, line 203: no need to give type, it's already inferred

src/components/input/bl-input.ts
```diff
@@ -261,22 +268,32 @@ export default class BlInput extends FormControlMixin(LitElement) {
...
+  }
+
   /**
    * Force to set input as in invalid state.
+   * @deprecated use error attribute instead
    */
   async forceCustomError() {
     await this.updateComplete;
-    this.validationTarget.setCustomValidity(this.customInvalidText || "An error occurred");
+    this.setCustomValidity(this.customInvalidText || "An error occurred");
```
AykutSarac, line 285: we have localized text here, either remove it or wrap with localization

src/utilities/form-control.test.ts
```diff
@@ -13,31 +13,58 @@ class MyInvalidInput extends LitElement {
   @query("input")
   validationTarget: HTMLInputElement;
 
+  @property({ reflect: true, type: String })
+  error: string = "";
```
AykutSarac, line 17: no need to give strict type, already inferred

src/utilities/form-control.test.ts
```diff
@@ -13,31 +13,58 @@ class MyInvalidInput extends LitElement {
...
   }
 }
 
+@customElement("my-custom-error-input")
+class MyCustomErrorInput extends LitElement {
+  @query("input")
+  validationTarget: HTMLInputElement;
+
+  @property({ reflect: true, type: String })
+  error: string = "";
```
AykutSarac, line 30: no need to give strict type

AykutSarac (changes requested):

src/components/input/bl-input.ts
Enes5519: It felt like a different context bro

src/components/input/bl-input.ts
Enes5519: I don't think this decorator is trained in TypeScript. It doesn't hurt to have it as an extra, bro.

AykutSarac: it's already inferred as part of TypeScript, no additional work required for decorator. since we don't have explicit types for other values it's better to remove it

src/components/input/bl-input.ts
AykutSarac: it'd be nice to solve it as we caught it, but fine to handle it in a separate pull request.

erbilnas (dismissed):

AykutSarac (approved):

PR #972, opened 2024-12-25
feat(link): add link component

### LINKED ISSUE #906: Link Component

erbilnas:
Closes #906 

- No breaking changes
- New component


src/baklava.ts
```diff
@@ -1,40 +1,41 @@
...
+export { default as BlTab } from "./components/tab-group/tab/bl-tab";
 export { default as BlTable } from "./components/table/bl-table";
-export { default as BlTableHeader } from "./components/table/table-header/bl-table-header";
 export { default as BlTableBody } from "./components/table/table-body/bl-table-body";
-export { default as BlTableRow } from "./components/table/table-row/bl-table-row";
-export { default as BlTableHeaderCell } from "./components/table/table-header-cell/bl-table-header-cell";
 export { default as BlTableCell } from "./components/table/table-cell/bl-table-cell";
-export { default as BlSplitButton } from "./components/split-button/bl-split-button";
-export { default as BlCalendar } from "./components/calendar/bl-calendar";
-export { default as BlDatePicker } from "./components/datepicker/bl-datepicker";
```
AykutSarac, line 39: why did we change casing? that could be a breaking change

src/components/link/bl-link.ts
```diff
@@ -0,0 +1,166 @@
...
+export default class BlLink extends LitElement {
+  static get styles(): CSSResultGroup {
+    return [style];
+  }
+
+  /**
+   * Target URL for the link
+   */
+  @property({ type: String, reflect: true })
+  target = "";
```
AykutSarac, line 30: add type safety, we can inherit from anchor's types

src/components/link/bl-link.ts
```diff
@@ -0,0 +1,166 @@
...
+   * Link kind - only applies to standalone variant
+   */
+  @property({ type: String, reflect: true })
+  kind: LinkKind = "primary";
+
+  /**
+   * Whether the link is external
+   */
+  @property({ type: Boolean, reflect: true })
+  external = false;
```
AykutSarac, line 54: do we need an `external` attribute here? adding an icon feels outside the boundaries of our design system, more like an UI kit addition

src/components/link/bl-link.ts
```diff
@@ -0,0 +1,166 @@
...
+   * Aria label for the link
+   */
+  @property({ type: String, attribute: "aria-label" })
+  ariaLabel = "";
+
+  /**
+   * Whether the link is disabled
+   */
+  @property({ type: Boolean, reflect: true })
+  disabled = false;
```
AykutSarac, line 66: links shouldn't have a disabled state

src/components/link/bl-link.ts
```diff
@@ -0,0 +1,166 @@
...
+        href="${ifDefined(this.disabled ? undefined : this.target)}"
+        class="${classMap(classes)}"
+        target="${this.external ? "_blank" : "_self"}"
+        rel="${ifDefined(this.external ? "noopener noreferrer" : undefined)}"
+        role="link"
+        aria-label="${ifDefined(this.ariaLabel || undefined)}"
+        aria-disabled="${this.disabled}"
+        tabindex="${this.disabled ? "-1" : "0"}"
+        @click="${this.handleClick}"
+        @keydown="${this.handleKeyDown}"
```
AykutSarac, line 153: do we need to handle them specifically, doesn't browser do it for us natively?

src/components/link/bl-link.ts
AykutSarac, line 1: add other [anchor element attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a)

AykutSarac (dismissed):

src/components/link/bl-link.ts
AykutSarac: don't add them by hand as much as possible, there are examples of inheriting existing types from elements

src/baklava.ts
erbilnas: It's breaking tests and btw it already named like that
`export default class BlDatepicker extends DatepickerCalendarMixin {`

erbilnas: We should talk with @dilandoogan

src/components/link/bl-link.ts
erbilnas: I structured it this way in case we wanted to add event handling in the future, but you’re right we can remove it for now

src/components/link/bl-link.ts
erbilnas: We should discuss

src/components/link/bl-link.ts
erbilnas: It's nice to see an icon for external sites to show people they will leave I think, let's discuss

src/components/link/bl-link.ts
erbilnas: The target here is not the anchor type but the href. I want to give it a more meaningful name

src/components/link/bl-link.ts
AykutSarac: are you possibly referring to the conversation above?
here I meant adding the other required attributes such as `rel` etc.

buseselvi (dismissed):

fatihhayri (approved):

Enes5519 (dismissed):

buseselvi (approved):

PR #998, opened 2025-01-29
fix(bl-calendar): Add RTL support for arrow icons to prevent visual s…

mehmetranas:
Issue
When the user selects RTL mode, the calendar icons are swapped incorrectly.

Fix
Explicitly set direction: ltr to ensure icons remain in the correct order.
Resolved incorrect icon placement while maintaining RTL layout.

### DIFF
--- a/src/components/calendar/bl-calendar.css
+++ b/src/components/calendar/bl-calendar.css
@@ -24,6 +24,7 @@
 
 .arrow {
   flex: 1;
+  direction: ltr;
 }
 
 .header-text {


erbilnas:
Hi @mehmetranas, I reviewed the changes in the playground, but I didn't observe any differences. Does this work for you?

mehmetranas:
Thanks @erbilnas for your notice! I’ve updated the RTL support logic—could you check again?

PR #1139, opened 2026-01-22
feat(input): add text-overflow ellipsis for disabled inputs

### LINKED ISSUE #607: [Feature Request]: Adding text-overflow:ellipsis to bl-input component

I think we need to add text-overflow: ellipsis to see three dots when text is too long for input component.
For example what I want to see:
Ekran Resmi 2023-05-31 13 11 47
but now I see this:
Ekran Resmi 2023-05-31 13 12 20

### Proposed solution

_No response_

### Alternatives considered

_No response_

_No response_

furkanbicak:
- Add text-overflow:ellipsis styling to disabled input elements
- Add white-space:nowrap and overflow:hidden for proper ellipsis display
- Add test case to verify ellipsis styling is applied
- Fixes #607

### DIFF
--- a/src/components/input/bl-input.css
+++ b/src/components/input/bl-input.css
@@ -150,6 +150,9 @@ input::-webkit-credentials-auto-fill-button {
 
 input:disabled {
   cursor: not-allowed;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  overflow: hidden;
 }
 
 input::-webkit-calendar-picker-indicator {
--- a/src/components/input/bl-input.test.ts
+++ b/src/components/input/bl-input.test.ts
@@ -784,5 +784,21 @@ describe("bl-input", () => {
         expect(el.validity.rangeOverflow).to.be.true;
       });
     });
+
+    describe("disabled input text overflow", () => {
+      it("should apply text-overflow ellipsis to disabled input", async () => {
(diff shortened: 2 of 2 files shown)


furkanbicak:
I have updated the branch with the latest changes from next. The implementation and tests are ready for review.

erbilnas (approved):

erbilnas (changes requested):
lint issues

furkanbicak:
Fixed lint issues. Ready for re-review 🙏 @erbilnas  @AykutSarac

Enes5519:
Thanks for the MR! I see the intent to handle long text, but I have some concerns about making ellipsis the default behavior for all disabled inputs.

When we force truncation, we prevent users from scrolling or selecting the text to see the full value. This isn't ideal for 'Best UX' because it hides data in a non-interactive state.

If a specific use case really requires truncation, the developer can still handle it externally by adding a title attribute or a custom Tooltip. However, as a Design System standard, we should keep the input's content fully legible by default to ensure accessibility and context.

furkanbicak:
Thanks for the clarification — that makes sense.

I agree that truncation shouldn’t be the default for disabled inputs from an accessibility/UX perspective. I’ll handle this at the application level when needed.

Appreciate the feedback!

PR #1143, opened 2026-02-01
Fix [BUG]: Style issue - BlAccordionGroup with single child #1084 Open

### LINKED ISSUE #1084: [BUG]: Style issue - BlAccordionGroup with single child
### Issue description
`BlAccordionGroup` with a single `BlAccordion` does not get rendered properly. There are some border issues.

I think `BlAccordionGroup` should also handle a single `BlAccordion`. There are some use cases where users add and remove accordion items dynamically, which can lead to `BlAccordionGroup`s with a single child.

#### Multiple BlAccordion ✅ 
Image

#### Single BlAccordion ❌ 
Image

### Baklava Version
No response

### Browser [e.g. chrome, safari]:
Chrome 

### Operating system

- OS: [e.g. iOS]:

- Any other details...

Low (slightly annoying)

FiratDede:
This PR fixes a styling bug reported in #1084 by [TaylanDurmaz](https://github.com/TaylanDurmaz) .
The fix ensures that when the first BIAccordion child in a BIAccordionGroup is not also the last child, the bottom border is removed by setting `--bl-accordion-border-bottom: 0.` As a result, this fix handles the border issue for BIAccordionGroup with single child because if BIAccordionGroup has only single  BIAccordionGroup child, its child bottom border would appear.
Fixes #1084

### DIFF
--- a/src/components/accordion-group/bl-accordion-group.css
+++ b/src/components/accordion-group/bl-accordion-group.css
@@ -7,6 +7,9 @@
 .accordion-group ::slotted(bl-accordion:first-child) {
   --bl-accordion-radius-bottom-right: 0;
   --bl-accordion-radius-bottom-left: 0;
+}
+
+.accordion-group ::slotted(bl-accordion:first-child:not(:last-child)) {
   --bl-accordion-border-bottom: 0;
 }
 


erbilnas:
The accordion group should be used when multiple accordions are required. For single use cases, you can use `<bl-accordion />` directly.

Issue #485, opened 2023-03-16
Focus state visual tests should be added


muratcorlu:
We should cover visual tests (storybook stories) for focus states for the components that has a focus state. And apparently we miss some:

- [ ] Switch component
- [ ] Button component
- [ ] Input component
- [ ] Select component
- [ ] Textarea component
- [ ] Radio component

Including those test will prevent us to accidentally break focus state indicators for them.

prakersh:
While building out these focus state stories, onUI could help during the manual verification step - it's a free, open-source browser extension that lets you annotate expected focus states directly on the rendered component in the browser. Useful for documenting what each focus state should look like before locking in the visual test baselines. https://github.com/onllm-dev/onUI

PR #1153, opened 2026-02-09
feat(button): add focus state story and optimize accessibility (#485)

irem-kaya:
Description: This PR addresses the missing visual tests for focus states as requested in #485. I have implemented a dedicated "Focus State" story for the bl-button component.

Key Changes:

Automation: Added a Storybook play function to automatically focus the button element on story load for consistent visual testing.

Accessibility (WCAG): Optimized color contrast (ratio 7.31:1) using specific foreground (#803C00) and hover colors to ensure compliance with WCAG 2.1 standards.

Validation: Verified with the Storybook Accessibility addon, resulting in 0 violations and 8 passes.
Closes part of #485


.vscode/settings.json
```diff
@@ -6,5 +6,39 @@
     "source.sortImports": "always",
     "source.removeUnusedImports": "always",
     "source.addMissingImports": "always"
-  }
```
erbilnas, line 9: we shouldn't push these remembered words. it should be stay as your local change.

src/components/button/bl-button.stories.mdx
```diff
@@ -54,7 +54,17 @@ import { Meta, Canvas, ArgsTable, Story } from '@storybook/addon-docs';
   }}
 />
 
-export const SingleButtonTemplate = (args) => html`<bl-button
```
erbilnas, line 57: instead of changing main template, it should be better creating a new button template to show focus stated

erbilnas (changes requested):
commented

irem-kaya:
I've refactored the code to maintain the original SingleButtonTemplate and moved accessibility-specific optimizations to the new FocusStateTemplate. I also ensured all stories use the template functions for consistency and removed local .vscode/settings.json changes. Ready for review!

irem-kaya:
Hi @erbilnas, I've addressed the requested changes. The accessibility styles are now only in the FocusStateTemplate and removed from other stories. The .vscode/settings.json is also removed from tracking. Could you please take a look when you have time? 🙏

Issue #1123, opened 2026-01-08
[Component]: Breadcrumb


buseselvi:
### Figma Design Document

https://www.figma.com/design/RrcLH0mWpIUy4vwuTlDeKN/Baklava-Design-Guide?node-id=30127-3068&t=2KatC0pHP0MvvtaW-1

### Attributes

| Attribute | Description | Default Value |
| --------------- | --------------- | --------------- |
| attr (`string`) | attr desc | - |

### Slots

| Name | Description | Default Content |
| --------------- | --------------- | --------------- |
| `default` slot | What to put | - |

### Events

| Event | Description |
| --------------- | --------------- |
| `bl-{event-name}` | Event description |

### CSS Custom Properties

| Property | Description | Default Value |
| --------------- | --------------- | --------------- |
| `--bl-name` | Property desc | - |

furkanbicak:
Hi 👋  
I couldn’t find an existing Breadcrumb component under `src/components`.
Before starting, could you please confirm whether:
- the component already exists under a different name, or
- this issue is intended as a placeholder for a future Breadcrumb implementation?

Thanks!

buseselvi:
Hi @furkanbicak 👋  this component is not implemented yet. I opened this issue to define a new design-ready component.

PR #1157, opened 2026-02-11
feat: #1123 breadcrumb component

gokhanakn:


buseselvi:
Great job @gokhanakn ! I just have 2 notes:
1. It jumps when clicking on the three dots, can we fix that?

https://github.com/user-attachments/assets/85e1cb42-9e26-49eb-9aec-37d3b90debcf

2. When the three dots open, can we use dropdown button's popover like in the design?

image
 
 Thank you 🚀

PR #1184, opened 2026-04-04
fix(react): replace React.lazy with synchronous createComponent in Re…

### LINKED ISSUE #1171: [BUG]: React wrappers use React.lazy unnecessarily, causing UI flashes
### Issue description

Every React component exported from @trendyol/baklava/dist/baklava-react is wrapped in React.lazy():

`var k = e.lazy(() => customElements.whenDefined("bl-button").then(() => ({
  default: t({ react: e, displayName: "BlButton", tagName: "bl-button", ... })
})));`

`React.lazy` always suspends on first render. This means any Baklava component triggers the nearest `<Suspense>` fallback the first time it mounts, causing a visible flash/blink of the entire page.

Impact:

- Most noticeable in dev mode especially big apps, (Vite serves modules individually, so customElements.whenDefined() takes longer to resolve), but technically happens in production too (1 frame)

- Every big app using Baklava with React needs to work around this — we had to build a hidden "preloader" component that pre-renders every Baklava component at startup just to warm up the lazy cache

Suggested fix:
The `customElements.whenDefined()` guard is unnecessary. Web components support progressive enhancement by design — you can put `<bl-button>` in the DOM before the custom element is registered, and it upgrades automatically once defined. The wrappers can be created synchronously:

```
// Before (causes suspension):
var k = e.lazy(() => customElements.whenDefined("bl-button").then(() => ({
  default: t({ react: e, ... })
})));

// After (no suspension):
var k = t({ react: e, displayName: "BlButton", tagName: "bl-button", ... });
```

We verified this by patching baklava-react.js directly - removing the React.lazy wrappers and using synchronous t() calls eliminated the flash completely

### Media & Screenshots

Blink/flash effect on each first render:

https://github.com/user-attachments/assets/964ababb-e193-4f05-b3c6-4a448f0f0c05

### Baklava Version

3.4.0-beta.5 (also tried with LTS)

### Operating system

_No response_

Medium (should be fixed soon)

- [x] This issue is directly related to React

Anexus5919:
React wrappers in `baklava-react.ts` were using `React.lazy()` + `customElements.whenDefined()` to create component wrappers. `React.lazy` **always** suspends on the first render, triggering the nearest `<Suspense>` fallback and causing a visible UI flash/blink even when the custom element is already registered.

This PR replaces the lazy pattern with direct ES module imports and synchronous `createComponent()` calls. The `@customElement` decorator in each component file registers the element as a side-effect of the import, guaranteeing `elementClass` is defined before `createComponent` accesses it.

### Before (causes suspension):
```ts
export const BlButton = React.lazy(() =>
  customElements.whenDefined("bl-button").then(() => ({
    default: createComponent({
      elementClass: customElements.get("bl-button"),
      ...
    })
  }))
);
```

### After (synchronous, no flash):
```ts
import BlButtonElement from "./components/button/bl-button";

export const BlButton = createComponent({
  elementClass: BlButtonElement,
  ...
});
```

### Benefits
- **No UI flash** - components render synchronously on first mount, no Suspense fallback triggered
- **No race conditions** - ES module evaluation order guarantees element registration before `createComponent()` runs
- **Tree-shakeable** - only imported components are bundled by the consumer
- **Simpler consumer API** - consumers no longer need a separate `import '@trendyol/baklava'` side-effect import when using React wrappers

### Changed files
- `cemPlugins/generateReactExports.js` - updated generator to emit direct imports and synchronous `createComponent()` calls
- `cemPlugins/utils/resolveFilePath.js` - extracted `resolveImportPath()` utility for raw import paths

## Checklist
- [ ] I have added test cases for this feature
- [ ] I have updated the documentation (if necessary)

Fixes #1171

### DIFF
--- a/cemPlugins/generateReactExports.js
+++ b/cemPlugins/generateReactExports.js
@@ -2,7 +2,7 @@ import { mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { join as pathJoin } from "node:path";
 import { pascalCase } from "pascal-case";
 import { format } from "prettier";
-import { resolveFilePath } from "./utils/resolveFilePath.js";
+import { resolveImportPath } from "./utils/resolveFilePath.js";
 import { resolveParsedType } from "./utils/resolveParsedType.js";
 
 const prettierConfig = JSON.parse(readFileSync(".prettierrc.json", "utf-8"));
@@ -26,16 +26,17 @@ export function generateReactExports() {
         throw new Error("Component not found!");
       }
 
-      const componentsCode = components
+      const resolved = components
         .map(([el, path]) => resolveComponent(el, path))
-        .filter(c => !!c[0])
-        .join("\n");
+        .filter(c => !!c);
+
+      const importStatements = resolved.map(c => c.importStatement).join("\n");
+      const componentBlocks = resolved.map(c => c.componentCode).join("\n");
 
       const code = `import React from "react";
 import { type EventName, createComponent } from "@lit-labs/react";
-
-type Constructor<T> = { new (): T };
-${componentsCode}`;
+${importStatements}
+${componentBlocks}`;
 
       const formattedCode = format(code, Object.assign(prettierConfig, { parser: "typescript" }));
       const outputPath = "./src";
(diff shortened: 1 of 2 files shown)


Anexus5919:
@Enes5519 @ergenekonyigit 
Kindly review this PR when you get a chance. Thanks!

Enes5519:
Great PR! We really appreciate the work put into fixing the issue with React.lazy causing the UI flash.
                                                                                                                                                                                                        
  Unfortunately, this cannot be merged as-is, as it will break the officially supported use case of loading baklava from a CDN via a <script> tag (such as jsdelivr). In this case, customElements.whenDefined()
  acts as the promise guaranteeing that elementClass is defined when called by createComponent. Without it, elementClass will simply be undefined when createComponent tries to use it.

  The actual problem isn’t with whenDefined(). React.lazy always suspends on the first render and triggers the Suspense fallback, regardless of how quickly the promise resolves. What we actually need is a solution
  that eliminates the flash without breaking async registration.

  We’re willing to look at other solutions that address both of these issues, but the current proposal would be a step back for users who load baklava from a CDN.

PR #1204, opened 2026-06-09
fix(popover): render in top layer to fix z-index/stacking issues (#1049)

### LINKED ISSUE #1049: [BUG]: Z-index issue for Tooltip component
### Issue description

When using bl-tooltip in bl-table-cell, it stays behind the row below.

### Media & Screenshots

Image

### Baklava Version

_No response_

### Operating system

- OS: [e.g. iOS]:

- Any other details...

Medium (should be fixed soon)

Enes5519:
## Summary

Fixes #1049. `bl-tooltip` inside a `bl-table-cell` rendered **behind** the next table row. The root cause is broader than the table: the popover (used by tooltip, datepicker, split-button) renders with `position: fixed; z-index` inside shadow DOM, so **any** ancestor stacking context (`transform`, `filter`, `opacity`, `overflow`, `position` + `z-index`) traps it — no `z-index` value can lift it out.

This PR makes `bl-popover` render in the browser **top layer** via the native [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API), which escapes all ancestor stacking contexts and overflow clipping. No consumer components were modified — they all use `bl-popover` and inherit the fix automatically.

## What changed

- **`bl-popover`** now adds `popover="manual"` to its floating element and calls `showPopover()` / `hidePopover()` on show/hide, promoting it to the top layer.
- UA `[popover]` defaults (`inset: 0; margin: auto; overflow: auto`) are reset so floating-ui's `left/top` positioning wins and the arrow isn't clipped.
- A feature-detection guard (`"popover" in HTMLElement.prototype`) preserves the previous `z-index`-based behavior as a graceful fallback on browsers without Popover API support. The existing JS dismiss logic (click-outside / Escape / single-open) is preserved by using `manual` (not `auto`).
- **`@floating-ui/dom` upgraded `1.5.3 → ^1.7.6`**: required so floating-ui correctly computes the containing block for top-layer elements. A `transform`/`filter` on an ancestor establishes the containing block for `position: fixed` even in the top layer; floating-ui >= 1.6 accounts for this, fixing popovers that otherwise rendered at the viewport origin (0,0) instead of next to their target.

## Why `manual` + top layer (not just a higher z-index)

z-index cannot escape an ancestor stacking context. The top layer is the only robust, standards-based mechanism that is immune to ancestor `transform`/`filter`/`overflow`/`z-index`. It keeps the element in the shadow DOM (no portal hack), so styles, slots, ARIA and events are preserved.

## Behavioral change (not a breaking API change)

No public attribute / property / event / slot signature changed. On supporting browsers, `--bl-popover-position` and `z-index` overrides become no-ops because the top layer ignores them (documented in the popover stories). On non-supporting browsers the old behavior is unchanged.

## Testing

- New tests: top-layer `:popover-open` state on show/hide, clean disconnect while open, a #1049 regression (tooltip in a table cell), and positioning inside a transformed ancestor.
- Full suite: **770 passed, 0 failed, 9 skipped** across Chromium / Firefox / WebKit, **coverage 100%**.
- Consumer suites (tooltip, datepicker, split-button, select) all pass — no consumer edits needed.
- Verified in the playground with real-browser measurements (Playwright): tooltip/popover positions correctly next to its target in a plain table cell, inside an `overflow + transform` container, and above a modal `bl-dialog` (which also uses the top layer).

## Notes / follow-ups (non-blocking, pre-existing)

- The unused `--bl-index-tooltip` token is now effectively obsolete for trapped-context cases; could be deprecated separately.
- Pre-existing: repeated `show()` reassigns the floating-ui `autoUpdate` cleanup without invoking the prior one; `disconnectedCallback` does not remove document listeners. Not worsened by this PR; candidates for a follow-up.
- The positioning regression test is a weak guard: web-test-runner's iframe isolation does not fully reproduce the real viewport/containing-block behavior, so it does not reliably catch this specific bug. The actual fix is the floating-ui upgrade, verified manually with Playwright.

Closes #1049

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0-dev",
       "license": "MIT",
       "dependencies": {
-        "@floating-ui/dom": "^1.2.3",
+        "@floating-ui/dom": "^1.7.6",
         "@fontsource-variable/rubik": "^5.0.2",
         "@lit-labs/react": "^2.0.3",
         "@lit/localize": "^0.12.1",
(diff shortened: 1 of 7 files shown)


okandavut (approved):
LGTM!

ergenekonyigit (approved):
LGTM

buseselvi (approved):

PR #1211, opened 2026-06-29
v3.5.0

Enes5519:
Stable release: next → main

## Highlights
- feat: add dark mode
- feat(popover): native Popover API top-layer rendering with fallback
- feat(datepicker): disabled clear icon when datepicker is disabled
- feat(i18n): add localization to pagination component
- fix(popover): correct top-layer positioning inside transformed ancestors
- fix(select): prevent duplicate option registration / incorrect unregistration
- fix(table): use MutationObserver for reactive row detection
- fix(deps): rename name to tag name for element class
- fix(deps): add event handlers to react export

Merge with a **merge commit** (not squash) so semantic-release can compute the version (expected `3.5.0`) and generate release notes from individual conventional commits.


ergenekonyigit (approved):
LGTM

src/components/pagination/bl-pagination.ts
fatihhayri: lint hataları gider

PR #1218, opened 2026-08-10
fix(button): keep hover styles when `disabled` is nullish

### LINKED ISSUE #728: [BUG]: Setting a property can disrupt the style
### Issue description

If we set the 'disabled' of the 'bl-button' to 'undefined,' it doesn't work in the hover state of secondary buttons because it [checks](https://github.com/Trendyol/baklava/blob/next/src/components/button/bl-button.css#L164) for 'aria-disabled=false'.

If you're wondering how to encounter this issue in real life, you can pass 'undefined' for 'disabled' in React to the 'BlButton' component; otherwise, setting the prop from JavaScript is sufficient.

### Media & Screenshots

_No response_

### Baklava Version

2.3.0

### Operating system

- OS: [e.g. iOS]: macOS
- Browser [e.g. chrome, safari]: Firefox

- Any other details...

Low (slightly annoying)

BilgeBalga:
`aria-disabled` is bound with `ifDefined()` over `loading || disabled`, so a nullish `disabled` — React's `disabled={undefined}`, or assigning the property from plain JS — resolves to `undefined` and drops the attribute from the DOM. The secondary and tertiary hover rules matched `[aria-disabled="false"]`, which then has nothing to match: the button stays interactive but loses its hover background and text color, leaving only the generic border change.

CSS: match `.button:not([aria-disabled="true"])` instead. A missing `aria-disabled` means "not disabled" per ARIA, so a single selector covers both the explicit `"false"` and the omitted case, and holds however the attribute is emitted. Keeping the fix in CSS rather than coercing `isDisabled` to a boolean leaves the rendered DOM unchanged for consumers and removes the brittle selector for good, instead of masking it at one call site.

Tests: drive a real pointer with `sendMouse` and assert computed styles, as `bl-tooltip` and `bl-notification-card` already do. Asserting against a hardcoded selector string would pass against the old CSS and guard nothing. Covers secondary, tertiary and the anchor variant with a nullish `disabled`, explicit `disabled={false}`, and negative cases for `disabled` and `loading`.

Closes #728

### DIFF
--- a/src/components/button/bl-button.css
+++ b/src/components/button/bl-button.css
@@ -145,12 +145,12 @@
   --main-color: transparent;
 }
 
-:host([variant="secondary"]:hover) .button[aria-disabled="false"] {
+:host([variant="secondary"]:hover) .button:not([aria-disabled="true"]) {
   --content-color: var(--bl-color-neutral-full);
   --bg-color: var(--main-hover-color);
 }
 
-:host([variant="tertiary"]:hover) .button[aria-disabled="false"] {
+:host([variant="tertiary"]:hover) .button:not([aria-disabled="true"]) {
   --content-color: var(--main-hover-color);
   --bg-color: var(--text-hover-color);
 }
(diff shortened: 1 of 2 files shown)


BilgeBalga:
Hi @Enes5519, I just re-triggered the GitHub Actions since the previous run had expired. Could you please approve the workflow run when you have a chance?

